PR #9, opened 2014-05-21
Disable the Active Directory Groups on Unix

rmessner:
This patch allow the end-user to disable the retrieving of groups from the Active Directory on Linux, directly from Security Configuration on Jenkins.


jtnord:
the timeout getting groups will be that the plugin is using MATCHING_RULE_IN_CHAIN (as this is done with auto)
If you change this to recursive (in the groups search) then you should not have the timeout.

jtnord:
timeout issues with group retrieval should have been fixed by @fbelzunc (not yet released) so I think the use case for this is no longer needed and can be closed.

fbelzunc:
Already released in 1.42

* https://issues.jenkins-ci.org/browse/JENKINS-33213
* https://issues.jenkins-ci.org/browse/JENKINS-28857

This should be already fixed.

PR #11, opened 2015-02-17
Improved multiple domain support

nieuwtje:
Whenever multiple domain controllers are defined, but they don't have a port assigned, only the last domain controller was given a port -> Added support for multiple servers without assigned ports.

Also, when multiple domain controllers are defined, only the ones in the domain you're looking at need to be checked -> Added check which checks whether hostname ends with the given domain, if not, it's not in the domain you're looking at.


jtnord:
can you split this into to separate PRs?

The not adding the port is perfectly valid and not contentiuos.  Filtering the preferred server is as this may result in not adding a preferred server that would be network local and serving the global catalogue (so login times could increase due to network RTT and/or quering multiple servers in the case of a bad password).

nieuwtje:
Moved port assignment to a new PR
> Whenever multiple domain controllers are defined, but they don't have a port assigned, only the last domain controller was given a port -> Added support for multiple servers without assigned ports.

jtnord:
So I'm not convinced that this is correct.  An AD server for sub.bob.com could be in foo.com and not sub.bob.com (e.g. a global catalog in the forest). If global.foo.com is site local and server.sub.bob.com is the other side of the world this would have a massive performance impact.

PR #12, opened 2015-02-17
[FIXED JENKINS-26737] AD can not log on with email address

nieuwtje:
Whenever someone tries to login with upn (someone@domain.com) the domain was appended to the username, this is unnecessary because the domain is already included in the upn.

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
@@ -369,16 +369,16 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
     /**
      * Returns the full user principal name of the form "joe@europe.contoso.com".
      * 
-     * If people type in 'foo@bar' or 'bar\\foo', it should be treated as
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -376,7 +376,7 @@ private String getPrincipalName(String username, String domainName) {
         if (slash>0) {
             principalName = username.substring(slash+1)+'@'+username.substring(0, slash)+'.'+domainName;
         } else if (username.contains("@"))
-            principalName = username+'.'+domainName;
+            principalName = username;
```
jtnord: this contradicts the javadoc behaviour `foo@bar -> foo@bar.acme.come`

nieuwtje: Updated the javadoc and the case where a username contains a backslash to match the desired pattern

jtnord:
:LGTM:
ping @fbelzunc

jtnord:
possibly related: [JENKINS-26737](https://issues.jenkins-ci.org/browse/JENKINS-26737)

PR #13, opened 2015-07-07
Fix findbugs issues

rsandell:
And configuration for xml output so Jenkins can pick it up

@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/AbstractActiveDirectoryAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/AbstractActiveDirectoryAuthenticationProvider.java
@@ -23,6 +23,7 @@
  */
 package hudson.plugins.active_directory;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.acegisecurity.AuthenticationException;
 import org.acegisecurity.providers.UsernamePasswordAuthenticationToken;
 import org.acegisecurity.providers.dao.AbstractUserDetailsAuthenticationProvider;
@@ -61,5 +62,6 @@ public abstract class AbstractActiveDirectoryAuthenticationProvider extends Abst
     /**
      * Setting this to true might help with diagnosing login problem.
      */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "It should, but changing would break backwards compatibility.")
     public static boolean SHOW_USER_NOT_FOUND_EXCEPTION = Boolean.getBoolean(AbstractActiveDirectoryAuthenticationProvider.class.getName()+".showUserNotFoundException");
 }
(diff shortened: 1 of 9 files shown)


Vlatombe:
How about adding the findbugs configuration and exclude filter to the plugins parent?

src/main/java/hudson/plugins/active_directory/AbstractActiveDirectoryAuthenticationProvider.java
```diff
@@ -61,5 +62,6 @@ protected void additionalAuthenticationChecks(UserDetails userDetails, UsernameP
     /**
      * Setting this to true might help with diagnosing login problem.
      */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "It should, but changing would break backwards compatibility.")
```
kohsuke: I don't agree that this "should" be final. These diagnostic fields are left mutable so that groovy console can be used to dynamically turn/off probes.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -676,6 +703,7 @@ protected UserDetails authenticate(String username, String password) throws Auth
      * @deprecated as of 1.28
      *      Use the UI field.
      */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "It should, but changing would break backwards compatibility.")
     public static String DOMAIN_CONTROLLERS = System.getProperty(ActiveDirectorySecurityRealm.class.getName()+".domainControllers");
```
kohsuke: Ditto as above

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -600,7 +609,25 @@ public DirContext createDNSLookupContext() throws NamingException {
...
 
                     public int compareTo(PrioritizedSocketInfo that) {
-                        return that.priority-this.priority; // sort them so that bigger priority comes first
+                        int prio = that.priority - this.priority; // sort them so that bigger priority comes first
+                        if (prio != 0) {
+                            return prio;
+                        }
+                        String s1 = socket != null ? socket.toString() : "";
+                        String s2 = that.socket != null ? that.socket.toString() : "";
+                        return s1.compareTo(s2);
```
kohsuke: :bug:  Sorting entries of the same priority by host name is unwise, as it forces all the clients to go through the same order and fails to balance the load. We should just stick to the priority sorting and the `socket` field shouldn't enter into the comparison.

hashCode and equal functions are dead code and not needed here, so I'd rather not have them for brevity, but maybe that's more of a matter of taste.

stephenc:
:bug: I agree with @kohsuke that the sorting should only be on the basis of priority as the information is supposed to be randomized within priority in order to even the load. Additionally the incorrect justification is given for the static non-final fields, the reason they are non-final is to allow run-time tuning via the groovy console. Citing the wrong justification is increasing technical debt

rsandell:
@Vlatombe sure, but that would take ages for most of the plugins to get to that parent version so they would need changes anyways.

One concern maybe would be that all plugins would then (once updated) get their findbugs warnings collected, most of them without any interested people looking at them and ending up wasting space on the ci server.

stephenc:
:bee:

rsandell:
@reviewbybees done

PR #14, opened 2015-07-28
Fixing open issue JENKINS-29280 which autofills passwords

dimacus:
Fixing this open bug https://issues.jenkins-ci.org/browse/JENKINS-29280

I could not find a simple Jelly solution to set the autocomplete attributes. Instead using a JS workaround to attach the autocomplete attribute after the page is loaded.

### DIFF
--- a/src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/configAdvanced.jelly
+++ b/src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/configAdvanced.jelly
@@ -22,4 +22,10 @@
   <f:nested>
     <f:validateButton with="domain,server,site,bindName,bindPassword" title="${%Test}" method="validate"/>
   </f:nested>
+	<script type="text/javascript" >
+		document.addEventListener("DOMContentLoaded", function(event) {
+		   document.getElementsByName("_.bindPassword")[0].autocomplete = "off";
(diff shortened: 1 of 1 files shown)


varmenise:
not sure but, would the jelly `default` option (set to ' ') work instead? https://wiki.jenkins-ci.org/display/JENKINS/Basic+guide+to+Jelly+usage+in+Jenkins#BasicguidetoJellyusageinJenkins-Defaultvalue.

dimacus:
Probably, sadly I'm not familiar with Jelly enough... would the default
overwrite any username/password that already there or it only make the
*default* happen when nothing else is pre-saved?

Also, would the browser still try to auto fill an empty field... and if you
put a ' ' to offset that, then the next commit will save a space for the
config. So we are back to square 1 :(

On Thu, Sep 24, 2015 at 2:36 PM, valentina <notifications@github.com> wrote:

> not sure but, would the jelly default option (set to ' ') work instead?
> https://wiki.jenkins-ci.org/display/JENKINS/Basic+guide+to+Jelly+usage+in+Jenkins#BasicguidetoJellyusageinJenkins-Defaultvalue
> .
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/active-directory-plugin/pull/14#issuecomment-143014151>
> .
>



-- 
-Dima Kovalenko
-------------------------

Good judgment comes from experience, and experience comes from bad judgment.
--Frederick P. Brooks

PR #17, opened 2015-11-17
[JENKINS-31575] Adding SSO/Client certificate authentication against

llecaroz:
# Context
Today, there are two plugins: the active-directory-plugin & the certificate-authentication-plugin
Both are great plugins.

But in big infras, administrators need SSO/Client certificate authentication against an Active Directory server. This means, that some administrators manage their groups in active directory but needs their end users to be authenticated by using client certificates.

# Features
This pull request will enable a such request (see this Jira ticket:  [JENKINS-31575 Adding SSO/Client certificate authentication against Active Directory](https://issues.jenkins-ci.org/browse/JENKINS-31575)
The SSO can be:
* Disabled, in this case the active directory will be the old one/AD workflow working alone
* SSO Enabled without check against Active directory. authenticated users will have a prefix set in their username and will be put in a default group
* SSO enabled with some matching rules (lower/upper case forcing/no change). In this case Bind DN username/password will be mandatory, the client certificate CN will be checked against the Active Directory username and its groups membership will be downloaded

# Development details
* **Help files available for all new options**
* **Upgrade in the POM to org.jenkins-ci.plugins:plugin:1.554 is mandatory due to the use of the jenkins.security.SecurityListener class* in SSOFilter.java**

This is a really must have feature which will be largely used by DevOps and in large companies.

Screenshot

# Reference (links)
* [Original certificate authentication plugin](https://github.com/jenkinsci/certificate-authentication-plugin)
* [Discussion](https://groups.google.com/forum/#!topic/jenkinsci-dev/Cpmq4fD67po)
* [Jira ticket](https://issues.jenkins-ci.org/browse/JENKINS-31575)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532</version>
+    <version>1.554</version>
   </parent>
 
   <artifactId>active-directory</artifactId>
@@ -109,4 +109,3 @@
     </license>
   </licenses>
 </project>
-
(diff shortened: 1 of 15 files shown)


llecaroz:
Note needed anymore, can be done with a standalone Filter plugin doFilter() by calling Jenkins.getInstance().getSecurityRealm().loadUserByUsername(username) without modifying the AD plugin and staying generic as most as possible.

PR #18, opened 2016-02-05
[FIXED JENKINS 16257] Active Directory Plugin - Credential exception tying to authenticate with special characters like / or #

fbelzunc:
We need to escape the special characters per:

* http://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx

@reviewbybees CC @jtnord

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
@@ -82,6 +82,8 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
 
     private GroupLookupStrategy groupLookupStrategy;
 
+    protected static final String DN_FORMATTED = "distinguishedNameFormatted";
+
     /**
      * {@link ActiveDirectoryGroupDetails} cache.
      */
@@ -299,25 +301,29 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
             }
             LOGGER.fine("Found user "+username+" : "+user);
 
-            Object dn = user.get("distinguishedName").get();
-            if (dn==null)
+            Object dnObject = user.get(DN_FORMATTED).get();
+            if (dnObject==null)
                 throw new AuthenticationServiceException("No distinguished name for "+username);
+
+            String dn = dnObject.toString();
+            LdapName ldapName = new LdapName(dn);
+            String dnFormatted = ldapName.toString();
 
             if (bindName!=null && password!=NO_AUTHENTICATION) {
                 // if we've used the credential specifically for the bind, we
                 // need to verify the provided password to do authentication
                 LOGGER.fine("Attempting to validate password for DN="+dn);
-                DirContext test = descriptor.bind(dn.toString(), password, ldapServers);
+                DirContext test = descriptor.bind(dnFormatted, password,ldapServers);
                 // Binding alone is not enough to test the credential. Need to actually perform some query operation.
                 // but if the authentication fails this throws an exception
                 try {
-                    new LDAPSearchBuilder(test,domainDN).searchOne("(& (userPrincipalName={0})(objectCategory=user))",userPrincipalName);
+                    new LDAPSearchBuilder(test,domainDN).searchOne("(& (userPrincipalName={0})(objectCategory=user))", userPrincipalName);
                 } finally {
                     closeQuietly(test);
                 }
             }
 
-            Set<GrantedAuthority> groups = resolveGroups(domainDN, dn.toString(), context);
+            Set<GrantedAuthority> groups = resolveGroups(domainDN, dnFormatted, context);
             groups.add(SecurityRealm.AUTHENTICATED_AUTHORITY);
 
             return new ActiveDirectoryUserDetail(username, password, true, true, true, true, groups.toArray(new GrantedAuthority[groups.size()]),
@@ -396,6 +402,9 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
      * @param context Used for making queries.
      */
     private Set<GrantedAuthority> resolveGroups(String domainDN, String userDN, DirContext context) throws NamingException {
+        if (userDN.contains("/")) {
+            userDN = userDN.replace("/","\\/");
+        }
         LOGGER.finer("Looking up group of "+userDN);
         Attributes id = context.getAttributes(userDN,new String[]{"tokenGroups","memberOf","CN"});
         Attribute tga = id.get("tokenGroups");
--- a/src/main/java/hudson/plugins/active_directory/LDAPSearchBuilder.java
+++ b/src/main/java/hudson/plugins/active_directory/LDAPSearchBuilder.java
@@ -70,11 +70,26 @@ class LDAPSearchBuilder {
         return this;
     }
 
+    /**
+     * Add all the attributes returned by the search
+     *
+     * @param filter
+     *      The filter to be used for the search
+     * @param args
+     *      Arguments to be used for the search
+     *
+     * @return Return all the attributes of the search and adds the DN correctly formatted per RFC 2253
+     */
     public Attributes searchOne(String filter, Object... args) throws NamingException {
         NamingEnumeration<SearchResult> r = search(filter,args);
         try {
             if (r.hasMore()) {
-                Attributes attrs = r.next().getAttributes();
+                SearchResult searchResult = r.next();
+                Attributes attrs = searchResult.getAttributes();
+                //We need to use getNameInNamespace in order to correctly format everything
+                //to be able to use LdapName later and get the DN correctly formatted with escaped
+                //characters like slash
+                attrs.put(ActiveDirectoryUnixAuthenticationProvider.DN_FORMATTED, searchResult.getNameInNamespace());
                 LOG.log(Level.FINER, "found {0}", attrs);
                 return attrs;
             } else {
--- a/src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-bindName.html
+++ b/src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-bindName.html
@@ -10,4 +10,8 @@
     <p>
     This field must be the full user principal name with domain name, like "joe@europe.contoso.com", or
     a LDAP-style distinguished name, such as "CN=Joe Chin,OU=europe,DC=contoso,DC=com".
+
+    <p>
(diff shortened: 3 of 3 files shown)


fbelzunc:
Forget to review this pull request. The issue still persist for some escaped characters

jtnord:
Bind expects a DN before a userPrincipalName.

msdn.microsoft.com/en-us/library/cc223499.aspx

Also a UPN is to my understanding not always present and requires a forest with a global catalog

jtnord:
> Also a UPN is to my understanding not always present and requires a forest with a global catalog

I read that somewhere this morning - but now I can't find the article; perhaps it was lack of :coffee: and I was misinterpreting a technet artical that said how to add the suffix to a UPN (and read that without the suffix!)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -299,25 +299,27 @@ public UserDetails retrieveUser(String username, String password, String domainN
             }
             LOGGER.fine("Found user "+username+" : "+user);
 
-            Object dn = user.get("distinguishedName").get();
-            if (dn==null)
+            Object userDistinguishedNameObject = user.get("distinguishedName").get();
```
jtnord: or just userDN to make it shorter :-)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -339,6 +341,74 @@ public UserDetails retrieveUser(String username, String password, String domainN
...
     }
 
+    /**
+     * Escape the DN escaped characters per http://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx
+     *
+     * @param distinguishedName
+     *      The distinguisedName without escaping the ten characters plus the forward slash
+     * @return never null
+     */
+    public static String escapeCharactersInDistinguisedName (String distinguishedName, boolean flagForwardSlash) {
```
jtnord: public?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -299,25 +299,27 @@ public UserDetails retrieveUser(String username, String password, String domainN
...
                 try {
-                    new LDAPSearchBuilder(test,domainDN).searchOne("(& (userPrincipalName={0})(objectCategory=user))",userPrincipalName);
+                    new LDAPSearchBuilder(test,domainDN).searchOne("(& (userPrincipalName={0})(objectCategory=user))", userPrincipalName);
                 } finally {
                     closeQuietly(test);
                 }
             }
-
-            Set<GrantedAuthority> groups = resolveGroups(domainDN, dn.toString(), context);
+            String userDistinguishedNameEscapingForwardSlash = escapeCharactersInDistinguisedName(userDistinguishedName, true);
```
jtnord: this escapes more than the forwardSlash.
I would also recommend that this is done inside resolveGroups and not here.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
jtnord: `flagForwardSlash`  why is this an option?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -339,6 +341,74 @@ public UserDetails retrieveUser(String username, String password, String domainN
...
+                            sb.append("\\;");
+                            break;
+                        case '"':
+                            sb.append("\"");
+                            break;
+                        case '=':
+                            sb.append("\\=");
+                            break;
+                        case ' ':
+                            sb.append("\\ ");
```
jtnord: this is incorrect - the Space **only** needs escaping if it is the [leading or trailing](http://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx) part of the DN.

> The space character must be escaped only if it is the leading or trailing character in any component of a distinguished name

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -339,6 +341,74 @@ public UserDetails retrieveUser(String username, String password, String domainN
...
+                            break;
+                        case '/':
+                            if (flagForwardSlash) {
```
jtnord: why?

src/test/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProviderTest.java
```diff
@@ -0,0 +1,56 @@
+package hudson.plugins.active_directory;
```
jtnord: (c) please

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-bindName.html
```diff
@@ -10,4 +10,8 @@
     <p>
     This field must be the full user principal name with domain name, like "joe@europe.contoso.com", or
     a LDAP-style distinguished name, such as "CN=Joe Chin,OU=europe,DC=contoso,DC=com".
+
+    <p>
+    Active Directory requires that the <a href="http://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx">following ten characters</a>
+    be escaped with the backslash "\" escape character if they appear in any of the individual components of a distinguished name.
```
jtnord: So this actually doesn't need to be a DN - it can also be a `UserPrincipalName` or `DOMAIN\user`

Slightly OT I know :-)

src/test/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProviderTest.java
```diff
@@ -0,0 +1,56 @@
...
+        distinguisedName = "CN=name=surname,OU=Jenkins-core-developer,DC=domain,DC=com";
+        assertEquals("CN=name\\=surname,OU=Jenkins-core-developer,DC=domain,DC=com", escapeCharactersInDistinguisedName(distinguisedName, false));
+
+        //Check escape of espace character
+        distinguisedName = "CN=name surname,OU=Jenkins-core-developer,DC=domain,DC=com";
+        assertEquals("CN=name\\ surname,OU=Jenkins-core-developer,DC=domain,DC=com", escapeCharactersInDistinguisedName(distinguisedName, false));
+
+        //Check escape of forward character
+        distinguisedName = "CN=name/surname,OU=Jenkins-core-developer,DC=domain,DC=com";
+        assertEquals("CN=name\\/surname,OU=Jenkins-core-developer,DC=domain,DC=com", escapeCharactersInDistinguisedName(distinguisedName, true));
```
jtnord: if you are keeping the option for escaping slash - you should verify with the flag == `false` that the escaping doesn't happen.

src/test/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProviderTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    public void testEscapeCharactersInDistinguisedName() {
+        //Check escape of comma character
+        String distinguisedName = "CN=name,surname,OU=Jenkins-core-developer,DC=domain,DC=com";
+        assertEquals("CN=name\\,surname,OU=Jenkins-core-developer,DC=domain,DC=com", escapeCharactersInDistinguisedName(distinguisedName, false));
+
+        //Check escape of backslash character
+        distinguisedName = "CN=name\\ surname,OU=Jenkins-core-developer,DC=domain,DC=com";
+        assertEquals("CN=name\\\\\\ surname,OU=Jenkins-core-developer,DC=domain,DC=com", escapeCharactersInDistinguisedName(distinguisedName, false));
```
jtnord: this is the wrong result.  
should be
`CN=name\\\\ surname,OU=Jenkins-core-developer,DC=domain,DC=com`

fbelzunc:
@jtnord Would you mind to review again, please?

fbelzunc:
@kohsuke Would you mind to review, please?

fbelzunc:
@reviewbybees would you mind to review again, please?

src/main/java/hudson/plugins/active_directory/LDAPSearchBuilder.java
```diff
@@ -70,11 +70,24 @@ public LDAPSearchBuilder returns(String... attributes) {
         return this;
     }
 
+    /**
+     * Add all the attributes returned by the search
+     *
+     * @param filter
```
varmenise: parameters should have a description

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-bindName.html
```diff
@@ -10,4 +10,8 @@
     <p>
     This field must be the full user principal name with domain name, like "joe@europe.contoso.com", or
     a LDAP-style distinguished name, such as "CN=Joe Chin,OU=europe,DC=contoso,DC=com".
+
+    <p>
+    In case you are using DN in the bind, Active Directory requires that the <a href="http://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx">following ten characters</a>
+    be escaped with the backslash "\" escape character if they appear in any of the individual components of a distinguished name.
```
varmenise: `are` escaped ?

varmenise:
Except those NITs the rest LGTM afaiu, so you have my :bee: but you would probably still want James and people with more expertise on the topic to look at it.

src/main/java/hudson/plugins/active_directory/LDAPSearchBuilder.java
```diff
@@ -70,11 +70,24 @@ public LDAPSearchBuilder returns(String... attributes) {
...
         NamingEnumeration<SearchResult> r = search(filter,args);
         try {
             if (r.hasMore()) {
-                Attributes attrs = r.next().getAttributes();
+                SearchResult searchResult = r.next();
+                Attributes attrs = searchResult.getAttributes();
+                //We need to use getNameInNamespace in order to correctly format everything
+                //to be able to use LdapName later and get the DN correctly formatted with escaped
+                //characters like slash
+                attrs.put("distinguishedNameFormatted", searchResult.getNameInNamespace());
```
jtnord: the naming makes this look a bit official - also as you use the same string elsewhere this should be a package scoped final variable.

jtnord:
a few comments inline - mostly a :bee:

PR #19, opened 2016-02-05
Added support for multiple servers without assigned ports

nieuwtje:
Whenever multiple domain controllers are defined, but they don't have a port assigned, only the last domain controller was given a port -> Added support for multiple servers without assigned ports.


jtnord:
code LGTM - but the history would be a little unclear due to the merge in the ancestry.  Can you rebase the change ontop of origin/master?

jtnord:
@fbelzunc FYI.

PR #24, opened 2016-03-02
[FIXED JENKINS-33260] Change the logic when resolving groups

fbelzunc:
```
if (!found || duration >= 10) {
                        LOGGER.warning(String.format("AD chain lookup is taking too long (%dms). Falling back to recursive lookup", duration));
                        groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
                        continue;
                    }
```

This logic smells wrong. We could have found the responses - just taken more than 10 seconds.

https://issues.jenkins-ci.org/browse/JENKINS-33260

@reviewbybees CC @jtnord

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
@@ -461,8 +461,16 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
                             throw e;
                         }
                     }
-                    if (!found || duration >= 10) {
-                        LOGGER.warning(String.format("AD chain lookup is taking too long (%dms). Falling back to recursive lookup", duration));
+                    if (!found && duration >= 10) {
+                        LOGGER.log(Level.WARNING, "Group lookup via Active Directory's 'LDAP_MATCHING_RULE_IN_CHAIN' extension timed out after {0} seconds. Falling back to recursive group lookup strategy for this and future queries", duration);
+                        groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
+                        continue;
+                    } else if (found && duration >= 10) {
+                        LOGGER.log(Level.WARNING, "Group lookup via Active Directory's 'LDAP_MATCHING_RULE_IN_CHAIN' extension matched user's groups but took {0} seconds to run. Switching to recursive lookup for future group lookup queries", duration);
+                        groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
+                        return groups;
+                    } else if (!found) {
+                        LOGGER.log(Level.WARNING, "Group lookup via Active Directory's 'LDAP_MATCHING_RULE_IN_CHAIN' extension failed. Falling back to recursive group lookup strategy for this and future queries");
                         groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
                         continue;
                     } else {


rsandell:
:bee:

varmenise:
:bee:

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -461,10 +461,14 @@ private String getPrincipalName(String username, String domainName) {
...
+                    if (!found && duration >= 10) {
+                        LOGGER.warning(String.format("AD chain didn't find the group in (%dms). Falling back to recursive lookup", duration));
                         groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
                         continue;
+                    } else if (found && duration >= 10) {
+                        LOGGER.warning(String.format("AD chain lookup found the group but took too long (%dms). Falling back to recursive lookup", duration));
+                        groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
+                        return groups;
                     } else {
                         // it run fast enough, so let's stick to it
```
varmenise: what if `(!found && duration <=10)`, looks like you go in the `else` and return `groups` even if we did not find them

varmenise:
> We could have found the responses - just taken more than 10 seconds.

Exactly, that's why we fall back to the recursive strategy. It's more about returning the found `groups`  in the first iteration if `found = true && duration>=10` that we were missing. So +1 to the change. But...
Have a concern:

`found = true && duration>=10` --> recursive , return groups
`found = false && duration>=10` --> recursive but do not return
`found = true && duration<10` --> chain, return groups
`found = false && duration<10` -->  ?

fbelzunc:
@varmenise I think you are right - it is better to cover the last case.

Thanks for carefully reviewing the pull request.

fbelzunc:
@varmenise Just covered that case.

Can you give a quick look again, please?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -461,8 +461,16 @@ private String getPrincipalName(String username, String domainName) {
                             throw e;
                         }
                     }
-                    if (!found || duration >= 10) {
-                        LOGGER.warning(String.format("AD chain lookup is taking too long (%dms). Falling back to recursive lookup", duration));
+                    if (!found && duration >= 10) {
+                        LOGGER.warning(String.format("AD chain didn't find the group in (%dms). Falling back to recursive lookup", duration));
```
jtnord: duration is in seconds no milliseconds

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -461,8 +461,16 @@ private String getPrincipalName(String username, String domainName) {
...
+                        groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
+                        continue;
+                    } else if (found && duration >= 10) {
+                        LOGGER.warning(String.format("AD chain lookup found the group but took too long (%dms). Falling back to recursive lookup", duration));
```
jtnord: duration is in seconds not milliseconds.
This is the wrong way to do [parameterized logging](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#log(java.util.logging.Level,%20java.lang.String,%20java.lang.Object))

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -461,8 +461,16 @@ private String getPrincipalName(String username, String domainName) {
...
+                        groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
+                        return groups;
+                    } else if (!found) {
+                        LOGGER.warning(String.format("AD chain lookup didn't found the group. Falling back to recursive lookup", duration));
```
jtnord: Nit: the English could be clearer.
"Group lookup via Active Directory\'s 'matching rule in chain' extension failed. Falling back to recursive group lookup strategy for this and future queries"

jtnord:
:bug: on the incorrect logging message (ms vs seconds)

varmenise:
:bee:

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -461,8 +461,16 @@ private String getPrincipalName(String username, String domainName) {
...
                         }
                     }
-                    if (!found || duration >= 10) {
-                        LOGGER.warning(String.format("AD chain lookup is taking too long (%dms). Falling back to recursive lookup", duration));
+                    if (!found && duration >= 10) {
+                        LOGGER.log(Level.WARNING, "Group lookup via Active Directory's 'LDAP_MATCHING_RULE_IN_CHAIN' extension timed out after {0} seconds. Falling back to recursive group lookup strategy for this and future queries", duration);
+                        groupLookupStrategy = GroupLookupStrategy.RECURSIVE;
+                        continue;
+                    } else if (found && duration >= 10) {
+                        LOGGER.log(Level.WARNING, "Group lookup via Active Directory's 'LDAP_MATCHING_RULE_IN_CHAIN' extension matched users groups but took {0} seconds to run. Switching to recursive lookup for future group lookup queries", duration);
```
jtnord: nit: user's

jtnord:
:bee:

PR #25, opened 2016-03-02
Correctly manage maven-release-plugin on pom.xml

fbelzunc:
@reviewbybees @jtnord

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,27 +72,26 @@
   </dependencies>
 
   <build>
-    <plugins>
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>findbugs-maven-plugin</artifactId>
-        <version>3.0.1</version>
-        <configuration>
-          <excludeFilterFile>findbugs-exclude.xml</excludeFilterFile>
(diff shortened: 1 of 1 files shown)


lordofthejars:
:bee:

pom.xml
```diff
@@ -88,9 +88,6 @@
         <groupId>org.apache.maven.plugins</groupId>
```
jtnord: this whole plugin section should go in `<pluginManagement>`

alecharp:
it's bad to have to do that but :bee: as it is because of the parent-pom configuration, not because of the plugin configuration.

jtnord:
you moved the findbugs - but not the m-r-p?

alecharp:
The two plugins were inside the `<pluginManagement/>` before. @fbelzunc followed my advices to move m-r-p outside of this section because of the parent pom

fbelzunc:
@alecharp Adding inside `pluginManagement` because otherwise when releasing you are not working inside the build lifecycle. Inside `pluginManagement` it will be valid inside build lifecycle and command line.

jtnord:
:bee:

PR #28, opened 2016-03-29
[FIXED JENKINS-21297] Optional cache for users and groups

fbelzunc:
Make cache optional for Users and Groups like in LDAP

I used the PR of LDAP plugin to [create this one](https://github.com/jenkinsci/ldap-plugin/commit/f3b01b2d4269a9a99f5880dd4406b4ebbfb74598)

https://issues.jenkins-ci.org/browse/JENKINS-21297

You still might need to cache the groups which an user is a member of which might be costly, but I will let this for a future PR.

@reviewbybees @jtnord


src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -99,7 +115,11 @@ static String dnToLdapUrl(String dn) {
         return "LDAP://"+dn.replace("/","\\/");
     }
 
-    protected UserDetails retrieveUser(String username, UsernamePasswordAuthenticationToken authentication) throws AuthenticationException {
+    /**
+     * Perform a look for and retrieve the UserDetails
+     * @return UserDetails
+     */
+    protected UserDetails computeUser(String username, UsernamePasswordAuthenticationToken authentication) throws AuthenticationException {
```
jtnord: nit: compute is really the wrong thing as it's not really computing - this is getting something from LDAP so `retreiveUserFromServer` may be more intuitive.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -68,6 +69,21 @@
...
 
+    /**
+     * To specify the TTL and Size used for caching users and groups
+     */
+    public CacheConfiguration cache;
+
+    /**
+     * {@link UserDetails} cache.
+     */
+    private transient Map<String,CacheConfiguration.CacheEntry<UserDetails>> userCache = null;
```
jtnord: Why not create these maps on desireialization/construction?
it would avoid the need for synchronization blocks on `this` and then use something like a `ConcurrentHashMap`

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -182,6 +187,23 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
         }
 
         this.server = server;
+        this.cache = cache;
+    }
+
+    public CacheConfiguration getCache() {
```
jtnord: @Restricted - don't want this API to leak to other plugins.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -143,6 +143,11 @@
      */
     public final boolean removeIrrelevantGroups;
 
+    /**
+     *  Cache of the Active Directory plugin
+     */
+    public CacheConfiguration cache;
```
jtnord: no to public fields.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -182,6 +187,23 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
 
         this.server = server;
+        this.cache = cache;
+    }
+
+    public CacheConfiguration getCache() {
+        return cache;
+    }
+
+    public void setCache(CacheConfiguration cache) {
```
jtnord: why public?

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+
+        private final int cacheSize;
+
+        public CacheMap(int cacheSize) {
+            super(cacheSize + 1); // prevent realloc when hitting cache size limit
+            this.cacheSize = cacheSize;
+        }
+
+        @Override
+        protected boolean removeEldestEntry(Map.Entry<K, CacheEntry<V>> eldest) {
```
jtnord: this does not ever seem to be called?

in which case just use a normal ConcurrentHashMap here and remove the syncronization

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param <K> Key type
+     * @param <V> Cache entry type
+     */
+    public static class CacheMap<K, V> extends LinkedHashMap<K, CacheEntry<V>> {
+
+        private final int cacheSize;
+
+        public CacheMap(int cacheSize) {
+            super(cacheSize + 1); // prevent realloc when hitting cache size limit
```
jtnord: I am missing something as nothing seems to control the size of the cache.  Things are just always added to it and are never removed.
so this will cache 1000000 groups if you ever look them up.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
jtnord: public without any restrictions?

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
jtnord: I think this should be initialized to a valid value (perhaps NO_CACHE) or something - but then the code does not have to check `if (cache != null) { /* do something */ }`

jtnord:
I don't think the cache is ever cleaned up (I may have missed it) which seems like a :bug: (either it should be cleaned up or the cleanup code should be removed and the cache simplified to remove locking etc..)
I also think bits are overly complicated as they have to check `(cache != null)` in various places - where the cache could be set to a valid value (no cache , 5 minute cache?) if one was not previously saved with the config.

naming police are also hanging on my shoulder about `computeXXX`

andresrc:
@fbelzunc use something like the Guava's cache (you have [this version](http://google.github.io/guava/releases/11.0.1/api/docs/com/google/common/cache/package-summary.html) on your classpath) instead of rolling your own. It works really well.

And as @jtnord said, hide whatever possible.

jtnord:
@andresrc you can't use guava here as the API is marked beta in the version used so the API can be broken (have not checked if it actually has been broken)

fbelzunc:
@jtnord 

https://github.com/jenkinsci/jenkins/blob/master/pom.xml#L185

```
<dependency>
        <groupId>com.google.guava</groupId>
        <artifactId>guava</artifactId>
        <version>11.0.1</version>
</dependency>
```

Why do you say this is beta?

jtnord:
Read [these notes](https://github.com/google/guava#important-warnings) then check the [API](https://github.com/google/guava/blob/v11.0/guava/src/com/google/common/cache/Cache.java#L49) you are intending to use...

fbelzunc:
@jtnord Latest guava release doesn't have the beta annotation on the cache.

I guess this might be a stupid question, but cannot we overwrite the version declared on the core in the plugin to the latest release?

* is not there a better solution other than continue with my own implementation?

jtnord:
> I guess this might be a stupid question, but cannot we overwrite the version declared on the core in the plugin to the latest release?

You can - and then you need to set the classloader to [plugin-first](https://wiki.jenkins-ci.org/display/JENKINS/Plugin+Structure#PluginStructure-Classloader). if Jenkins leaks guava though any API  you are a calling (or you leak it via any API) then you may be about to enter a world of pain.

andresrc:
@fbelzunc no need to upgrade the dependency, it is not worth the effort, and better to stick with the version you get from core.

The `Cache` API got out of beta in version 19  but the parts you need did not experiment any change (I've been using it extensively since the first releases) and, as far as I can tell (will check again in the review) are binary compatible, so from my POV you are good to go. In any case check that the APIs you use (mainly the `CacheBuilder` and `Cache.get(K, Callable)`) are in both versions 11 and 19.

/cc @jtnord

fbelzunc:
Still work in progress. There are several bugs need to be fixed. 

I will ping you once this is not a WiP anymore.

fbelzunc:
@andresrc @jtnord  I would like you review this.

I found two problems I am not sure how to fix.

1. When using `ActiveDirectoryAuthenticationProvider` - Windows - you don't have the realm so cannot read `CacheConfiguration` - not sure how to make the size and ttl configurable here.
2. In `ActiveDirectoryUnixAuthenticationProvider` I needed to build the cache in the constructor - it might be wrong but I checked the object is only created when saving the Global Security Configuration and under a Jenkins start - not sure if it is wrong but it looks wrong to me, although I am not sure how to workaround the problem.

Any ideas?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -99,64 +124,84 @@ static String dnToLdapUrl(String dn) {
...
+                        COM4J.removeListener(col);
+                    }
+                }
+            });
+        } catch (UncheckedExecutionException e) {
+            Throwable t = e.getCause();
+            if (t instanceof AuthenticationException) {
+                AuthenticationException authenticationException = (AuthenticationException)t;
+                throw authenticationException;
+            }
```
jtnord: Still need to throw an `AuthenticationException` to signal a failed login though don't you?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -99,64 +124,84 @@ static String dnToLdapUrl(String dn) {
...
+        } catch (java.util.concurrent.ExecutionException e) {
+            LOGGER.log(Level.SEVERE, "There was a problem caching user "+ username, e);
```
jtnord: Still need to throw an AuthenticationException to signal a failed login though don't you?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -99,64 +124,84 @@ static String dnToLdapUrl(String dn) {
...
         }
+        return userDetails;
```
jtnord: according to the [latest spring docs](http://docs.spring.io/autorepo/docs/spring-security/3.2.6.RELEASE/apidocs/index.html?org/springframework/security/authentication/dao/AbstractUserDetailsAuthenticationProvider.html) (which may have changed since acegi) this must never be `null` - yet you have code paths where this could be the case....

andresrc:
@fbelzunc You can have the caches in the realm and use them from the providers. That way they are "recreated" (which involves "purging") in every configuration change.

By default they can be configured or to a conservative value (even as conservative as no cache).

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -68,6 +73,26 @@
...
+    /**
+     * The {@link UserDetails} cache.
+     *
+     * By default set-up with size 1000 and ttl 3600 for legacy reasons
+     */
+    Cache<String, UserDetails> userCache = CacheBuilder.newBuilder()
+            .maximumSize(1000)
+            .expireAfterWrite(3600, TimeUnit.SECONDS)
+            .build();
+
```
andresrc: These default values don't match those in the help, and those ones seem more sensible.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -68,6 +73,26 @@
...
+    /**
+     * The {@link ActiveDirectoryGroupDetails} cache.
+     *
+     * By default set-up with size 1000 and ttl 3600 for legacy reasons
+     */
+    Cache<String, ActiveDirectoryGroupDetails> groupCache = CacheBuilder.newBuilder()
+            .maximumSize(1000)
+            .expireAfterWrite(3600, TimeUnit.SECONDS)
+            .build();
+
```
andresrc: ditto

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class CacheConfiguration<K,V,E extends Exception> {
+    public final int size;
+    public final int ttl;
+
+    @DataBoundConstructor
+    public CacheConfiguration(int size, int ttl) {
+        this.size = Math.max(10, Math.min(size, 1000));
+        this.ttl = Math.max(30, Math.min(ttl, 3600));
+    }
```
andresrc: No way to disable?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -250,109 +307,201 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
...
-                throw new UserMayOrMayNotExistException("Unable to retrieve the user information without bind DN/password configured");
-            }
+                        String dn = dnObject.toString();
+                        LdapName ldapName = new LdapName(dn);
+                        String dnFormatted = ldapName.toString();
+
+                        if (bindName != null && !password.equals(NO_AUTHENTICATION)) {
+                            // if we've used the credential specifically for the bind, we
+                            // need to verify the provided password to do authentication
+                            LOGGER.fine("Attempting to validate password for DN=" + dn);
```
jtnord: Don't perform string concatenation in conditional logging statements.
Replace with `LOGGER.log(Level.FINE, "blah {}", dn);` or whatever the correct idoyim is for parameterized logging in the logging framework that is being used.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -250,109 +307,201 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
...
-            throw new BadCredentialsException("Failed to retrieve user information for "+username, e);
-        } finally {
-            closeQuietly(context);
+                        Set<GrantedAuthority> groups = resolveGroups(domainDN, dnFormatted, context);
+                        groups.add(SecurityRealm.AUTHENTICATED_AUTHORITY);
+
+                        return new ActiveDirectoryUserDetail(username, password, true, true, true, true, groups.toArray(new GrantedAuthority[groups.size()]),
+                                getStringAttribute(user, "displayName"),
+                                getStringAttribute(user, "mail"),
+                                getStringAttribute(user, "telephoneNumber")
```
jtnord: what on earth are we doing with the telephone number?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -250,109 +307,201 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
...
+                        if (anonymousBind && e.getMessage().contains("successful bind must be completed") && e.getMessage().contains("000004DC")) {
+                            // sometimes (or always?) anonymous bind itself will succeed but the actual query will fail.
+                            // see JENKINS-12619. On my AD the error code is DSID-0C0906DC
+                            throw new UserMayOrMayNotExistException("Unable to retrieve the user information without bind DN/password configured");
+                        }
+
+                        LOGGER.log(Level.WARNING, "Failed to retrieve user information for " + username, e);
+                        throw new BadCredentialsException("Failed to retrieve user information for " + username, e);
+                    } catch (AuthenticationException e) {
+                        throw e;
```
jtnord: why catch and rethrow?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -250,109 +307,201 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
...
+                }
+            });
+        } catch (UncheckedExecutionException e) {
+           Throwable t = e.getCause();
+            if (t instanceof AuthenticationException) {
+                AuthenticationException authenticationException= (AuthenticationException)t;
+                throw authenticationException;
+            }
+        } catch (ExecutionException e) {
+            LOGGER.log(Level.SEVERE, "There was a problem caching user "+ username, e);
```
jtnord: needs to throw something - must never return `null`.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -250,109 +307,201 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
...
+        }
+        // We need to check the password when the user is cached so it doesn't get automatically authenticated
+        // without verifying the credentials
+        if (password != null && !password.equals(userDetails.getPassword())) {
+            throw new BadCredentialsException("Failed to retrieve user information from the cache for "+ username);
         }
+        return userDetails;
```
jtnord: potential `null` return to something that must never be `null`.

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public int getSize() {
+        return size;
+    }
+
+    public int getTtl() {
```
jtnord: Javadoc...
units.  milliseconds, seconds, minutes....

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+    public final int size;
+    public final int ttl;
+
+    @DataBoundConstructor
+    public CacheConfiguration(int size, int ttl) {
+        this.size = Math.max(10, Math.min(size, 1000));
+        this.ttl = Math.max(30, Math.min(ttl, 3600));
+    }
+
+    public int getSize() {
```
jtnord: javadoc

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,25 @@
+package hudson.plugins.active_directory;
+
+import org.kohsuke.stapler.DataBoundConstructor;
+
+/**
+ * Cache for Users and Groups
+ */
+public class CacheConfiguration<K,V,E extends Exception> {
+    public final int size;
```
jtnord: `private`

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+    public final int ttl;
```
jtnord: `private`

If you ever find you have written a public field that is not `final static` then you should stop and think "why?"

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @DataBoundConstructor
+    public CacheConfiguration(int size, int ttl) {
```
jtnord: Javadoc

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,25 @@
+package hudson.plugins.active_directory;
+
+import org.kohsuke.stapler.DataBoundConstructor;
+
+/**
+ * Cache for Users and Groups
```
jtnord: Nope - this is not a cache.  It is the configuration for a cache.

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-cacheTtl.html
```diff
@@ -0,0 +1,8 @@
+<div>
+    Some LDAP servers may be slow, or rate limit client requests. In such cases enabling caching may improve performance
+    of Jenkins with the risk of delayed propagation of user changes from LDAP and increased memory usage on the master.
+    <br/>
+    TTL will determinate the time the information will be cached
+    <br/>
+    <b>Note:</b> The default configuration is to have TTL to 10 minutes
```
jtnord: what units is this measured in?  seconds, minutes, hours, days, average lifespan of an adult mayfly?

The dafult may be 10 minutes - but once you have changed this to something other than 10 you have no idea anymore

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-cacheSize.html
```diff
@@ -0,0 +1,10 @@
+<div>
+    <div>
```
jtnord: this is generic to the cache - so can this be pulled into the outer help - and leave this just to the specific of the size?

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-cacheTtl.html
```diff
@@ -0,0 +1,8 @@
+<div>
+    Some LDAP servers may be slow, or rate limit client requests. In such cases enabling caching may improve performance
```
jtnord: this is generic to the cache - so can this be pulled into the outer help - and leave this just to the specific of the size?

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-cacheTtl.html
fbelzunc: This is specified on the `doFill..` but in any case I will clarify it here.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
fbelzunc: Because later you need to catch it in the next catch `UncheckedExecutionException` to re-throw again so the login is failed

```
catch (UncheckedExecutionException e) {
           Throwable t = e.getCause();
            if (t instanceof AuthenticationException) {
                AuthenticationException authenticationException= (AuthenticationException)t;
                throw authenticationException;
            } else {
                throw new GuavaAuthenticationException("Authentication failed because there was a problem caching user " +  username);
            }
        }
```

Do you suggest a different alternative?

jtnord: don't catch it to begin with?

fbelzunc:
@andresrc @jtnord You can review again.

I think it should now close to the right solution.

I am worried for creating the cache in the constructor of the class. Would you mind to review specially this part, please?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -68,7 +71,21 @@
      */
     private final _Connection con;
 
+    /**
+     * The {@link UserDetails} cache.
+     */
+    Cache<String, UserDetails> userCache;
```
andresrc: better `private` (and `final`)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -68,7 +71,21 @@
...
+
+    /**
+     * The {@link ActiveDirectoryGroupDetails} cache.
+     */
+    Cache<String, ActiveDirectoryGroupDetails> groupCache;
```
andresrc: better `private` (and `final`)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -99,63 +121,80 @@ static String dnToLdapUrl(String dn) {
...
+                    }
+                }
+            });
+        } catch (UncheckedExecutionException e) {
+            Throwable t = e.getCause();
+            if (t instanceof AuthenticationException) {
+                AuthenticationException authenticationException = (AuthenticationException)t;
+                throw authenticationException;
+            } else {
+                throw new GuavaAuthenticationException("Authentication failed because there was a problem caching user " + username);
```
andresrc: I'd go for the existing `AuthenticationException` one with the caught one as the cause.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -99,63 +121,80 @@ static String dnToLdapUrl(String dn) {
...
-            ).updateUserInfo();
-        } catch (AuthenticationException e) {
-            LOGGER.log(Level.FINE, "Failed toretrieve user "+username, e);
-            throw e;
-        } finally {
-            col.disposeAll();
-            COM4J.removeListener(col);
+        } catch (java.util.concurrent.ExecutionException e) {
+            LOGGER.log(Level.SEVERE, String.format("There was a problem caching user %s", username), e);
+            throw new GuavaAuthenticationException("Authentication failed because there was a problem caching user " + username);
```
andresrc: See above.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -223,42 +262,50 @@ private String getDnOfUserOrGroup(String userOrGroupname) throws UsernameNotFoun
...
+            if (t instanceof AuthenticationException) {
+                AuthenticationException authenticationException = (AuthenticationException)t;
+                throw authenticationException;
+            } else {
+                throw new GuavaAuthenticationException("Authentication failed because there was a problem caching group " +  groupname);
             }
+        } catch (java.util.concurrent.ExecutionException e) {
+            LOGGER.log(Level.SEVERE, String.format("There was a problem caching group %s", groupname), e);
+
+            throw new GuavaAuthenticationException("Authentication failed because there was a problem caching group " +  groupname);
```
andresrc: See above.

src/main/java/hudson/plugins/active_directory/GuavaAuthenticationException.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * To throw AuthenticationException when the login failed due a cache issue
+ */
+public class GuavaAuthenticationException extends AuthenticationException {
+
+    public GuavaAuthenticationException(String msg) {
+        super(msg);
+    }
+}
```
andresrc: As said above I would not use this class but the existing with a cause.

src/main/java/hudson/plugins/active_directory/CacheConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Get TTL of the cache
+     *
+     * @return the ttl of the cache in seconds
+     */
+    public int getTtl() {
+        return ttl;
+    }
+}
```
andresrc: I would move the cache construction here (using a static factory method, so that you can account on the null and easily type-inference the generics)

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -153,13 +163,33 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
     public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
                                         String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups) {
-        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups,domain!=null);
+        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups,domain!=null, null);
     }
+
+    /**
+     * CacheBuilder for users
+     */
+    CacheBuilder userCacheBuilder = CacheBuilder.newBuilder();
```
andresrc: Not needed, see below.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -153,13 +163,33 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
+
+    /**
+     * CacheBuilder for groups
+     */
+    CacheBuilder groupCacheBuilder = CacheBuilder.newBuilder();
```
andresrc: ditto

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -153,13 +163,33 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
+
+    /**
+     * The cache for users
+     */
+    Cache userCache;
+
```
andresrc: `private final`

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -153,13 +163,33 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
+    /**
+     * The cache for groups
+     */
+    Cache groupCache;
```
andresrc: `private final`

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -182,6 +212,55 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
+                    .expireAfterWrite(600, TimeUnit.SECONDS)
+                    .build();
+
+            groupCache = groupCacheBuilder
+                    .maximumSize(256)
+                    .expireAfterWrite(600, TimeUnit.SECONDS)
+                    .build();
+        }
+    }
+
```
andresrc: Move all this logic to the `CacheConfiguration`, as recommended below.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -182,6 +212,55 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
+    @Restricted(NoExternalUse.class)
+    private CacheConfiguration getCache() {
+        return cache;
+    }
+
+    @Restricted(NoExternalUse.class)
+    private void setCache(CacheConfiguration cache) {
+        this.cache = cache;
+    }
+
```
andresrc: Are those methods needed when you have the field? Can the value change after construction? Can the field be `private` (if only filled through the `DataBoundConstructor`)?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -86,63 +92,20 @@
...
-                try {
-                    DirContext context = descriptor.bind(bindName, bindPassword, obtainLDAPServers(domainName));
+    private CacheConfiguration cache;
 
-                    try {
-                        final String domainDN = toDC(domainName);
+    /**
+     * The {@link UserDetails} cache.
+     */
+    Cache<String, UserDetails> userCache;
```
andresrc: See above.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -86,63 +92,20 @@
...
-                    LOGGER.log(Level.WARNING, "Failed to find the group "+groupname+" in "+domainName+" domain", e);
-                    problem = true;
-                } finally {
-                    Thread.currentThread().setContextClassLoader(ccl);
-                }
-            }
+    /**
+     * The {@link ActiveDirectoryGroupDetails} cache.
+     */
+    Cache<String, ActiveDirectoryGroupDetails> groupCache;
```
andresrc: See above.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -152,51 +115,138 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
...
+            userCache = CacheBuilder.newBuilder()
+                    .maximumSize(256)
+                    .expireAfterWrite(600, TimeUnit.SECONDS)
+                    .build();
+
+            groupCache = CacheBuilder.newBuilder()
+                    .maximumSize(256)
+                    .expireAfterWrite(600, TimeUnit.SECONDS)
+                    .build();
+        }
```
andresrc: See above.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -250,109 +300,202 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
...
+                            }
+                        }
+                    });
+        } catch (UncheckedExecutionException e) {
+            Throwable t = e.getCause();
+            if (t instanceof AuthenticationException) {
+                AuthenticationException authenticationException= (AuthenticationException)t;
+                throw authenticationException;
+            } else {
+                throw new GuavaAuthenticationException("Authentication failed because there was a problem caching group " +  groupname);
```
andresrc: See above.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
fbelzunc: @andresrc The problem is that `AuthenticationException ` is abstract.

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-cacheSize.html
fbelzunc: I think I address your concern.

fbelzunc:
@andresrc @jtnord I think I addressed all the comments :-)

fbelzunc:
Squashing everything to a single commit and doing some testing.

fbelzunc:
@jtnord @andresrc 

By default, the cache will be disabled.

I tested and added a couple of commits more to fix a couple of problem I found. Now, under my point of view, it should be fine to merge now.

Please, do a quick review when you have the chance.

PR #29, opened 2016-04-11
[FIXED JENKINS-34143] Test Active Directory connection button reports success if the search operation doesn't have any result

fbelzunc:
When testing the AD connection details you need to actually do a search to make sure the credential is valid.
https://github.com/jenkinsci/active-directory-plugin/blob/master/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L392
If in this search there are not elements, we are returning null.
https://github.com/jenkinsci/active-directory-plugin/blob/master/src/main/java/hudson/plugins/active_directory/LDAPSearchBuilder.java#L98
And in this case we are saying that everything is success.
https://github.com/jenkinsci/active-directory-plugin/blob/master/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L425
it might happen that under AD replication farm one of the AD instances is broken not returning any user in the search. In this case, we should expose the issue on the GUI.

https://issues.jenkins-ci.org/browse/JENKINS-34143

@reviewbybees @escoem @andresrc @jtnord

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -389,7 +389,10 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
                             DirContext context = bind(bindName, Secret.toString(password), servers);
                             try {
                                 // actually do a search to make sure the credential is valid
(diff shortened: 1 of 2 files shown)


jtnord:
:bee: good catch.

andresrc:
:bee: :+1:

fbelzunc:
@reviewbybees done

PR #30, opened 2016-04-11
Add readme

fbelzunc:
@reviewbybees esp @jtnord @escoem

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,6 @@
+Active Directory Plugin
+
+See wiki at: https://wiki.jenkins-ci.org/display/JENKINS/Active+Directory+plugin 
+
+[![Build Status](https://jenkins.ci.cloudbees.com/job/plugins/job/active-directory-plugin/badge/icon)](https://jenkins.ci.cloudbees.com/job/plugins/job/active-directory-plugin/)
+


jtnord:
whilst I'm ok with this - why not document the trouble shooting in the wiki and point this page to the wiki for more information.   The wiki is normally looked at by users, and the repo by developers - and this information is usable by users!

README.md
```diff
@@ -0,0 +1,70 @@
+active-directory-plugin
```
oleg-nenashev: NIT: could be replaced by a human-readable name

README.md
```diff
@@ -0,0 +1,70 @@
...
+================
+
+[![Build Status](https://jenkins.ci.cloudbees.com/job/plugins/job/active-directory-plugin/badge/icon)](https://jenkins.ci.cloudbees.com/job/plugins/job/active-directory-plugin/)
```
oleg-nenashev: Such embedded status is useless till the PR builder runs within the same job. Better to remove till we have multibranch

oleg-nenashev:
Weak :-1: , because it causes the documentation duplication between Wiki and GtHub. Two instances have different info inside. My recommendation is to have docs in a single place

fbelzunc:
@reviewbybees done

PR #31, opened 2016-04-26
[FIXED JENKINS-34444] Test button is reporting managerDN binding is successful but was not able to find any user on the tree

fbelzunc:
In case in the root tree there are only OU and any USER, then the ad test button will report this. It is better to use a subtree scope so we actually look inside OUs.

https://issues.jenkins-ci.org/browse/JENKINS-34444

@reviewbybees CC @jtnord


jtnord:
:bee:

fbelzunc:
@reviewbybees done

PR #33, opened 2016-04-29
[JENKINS-34508] Migrate to 2.7 parent pom

armfergom:
- [x] Migrate to parent pom 2.7
- [x] Fix Findbugs and Animal Sniffer errors
- [x] Fix tests

@reviewbybees

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build
 .project
 .settings
 /work/
+bin
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
   </developers>
 
   <properties>
-    <jenkins.version>1.580.1</jenkins.version>
+    <jenkins.version>1.554.1</jenkins.version>
     <java.level>7</java.level>
     <powermock.version>1.6.1</powermock.version>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532</version>
+    <version>2.7</version>
   </parent>
 
   <artifactId>active-directory</artifactId>
@@ -32,7 +32,9 @@
   </developers>
 
   <properties>
-    <powermock.version>1.5.5</powermock.version>
+    <jenkins.version>1.580.1</jenkins.version>
+    <java.level>7</java.level>
+    <powermock.version>1.6.1</powermock.version>
   </properties>
 
   <dependencies>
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -32,7 +32,9 @@
   </developers>
 
   <properties>
-    <powermock.version>1.5.5</powermock.version>
+    <jenkins.version>1.580.1</jenkins.version>
```
fbelzunc: Now, minimum version is `1.580.1`? It was before 1.532 - this look wrong to me. Doesn't it?

What is the reason?

.gitignore
```diff
@@ -6,3 +6,4 @@ build
 .project
 .settings
 /work/
+bin
```
fbelzunc: why `bin` is needed?

pom.xml
armfergom: With the upgrade to the new test harness, tests start failing with the following error:

> testNoGroupsAreRegistered(hudson.plugins.active_directory.RemoveIrrelevantGroupsTest)  Time elapsed: 0 sec  <<< ERROR!
java.lang.NoClassDefFoundError: jenkins/model/DirectlyModifiableTopLevelItemGroup
	at hudson.plugins.active_directory.RemoveIrrelevantGroupsTest.<init>(RemoveIrrelevantGroupsTest.java:56)
Caused by: java.lang.ClassNotFoundException: jenkins.model.DirectlyModifiableTopLevelItemGroup
	at hudson.plugins.active_directory.RemoveIrrelevantGroupsTest.<init>(RemoveIrrelevantGroupsTest.java:56)

Looking at [DirectlyModifiableTopLevelItemGroup](https://github.com/jenkinsci/jenkins/blob/9fce1ee933eb5276baff977d562fc8e183f1c8d6/core/src/main/java/jenkins/model/DirectlyModifiableTopLevelItemGroup.java), you can see that it exists since `1.548` (which is newer than `1.533`). To avoid the error above, we need `1.548+`, I have chosen `1.580.1` because it is an LTS and includes SECURITY-144

.gitignore
armfergom: Eclipse auto-generates that folder with all compiled sources and resources when imported and I don't want to push that or for it to pollute my uncommitted changes. Happy to remove it if you want though, since I can always just delete the folder :)

pom.xml
fbelzunc: Can you change this to the next LTS from `1.548+`?

armfergom: Sure. Does `1.554.3` works for you?

fbelzunc:
Agreed that it will be done agains `1.554.1` and wait 1 week to see there is any objection for someone else.

armfergom:
@fbelzunc dropped baseline to 1.554.1

fbelzunc:
Thank you - I will wait one week to see if there is any objection and then merge.

armfergom:
@fbelzunc Just noticed that `<java.level>` must be set to 6 for that baseline and pushed the change. (See [README](https://github.com/jenkinsci/plugin-pom) of plugin-pom for guidelines to select java.level for different baselines.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java
```diff
@@ -115,7 +117,7 @@ public int hashCode() {
         return result;
     }
 
-    @Override
+    @Override @SuppressFBWarnings(value="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE", justification="https://github.com/jenkinsci/jenkins/pull/2094")
```
recena: Line long enough as to define two annotations in the same line.

.gitignore
recena: +1 to include it.

recena:
🐝

andresrc:
:bee:

armfergom:
@reviewbybees done

PR #34, opened 2016-05-17
[JENKINS-34426] Fix to handle SECURITY-243

fbelzunc:
# Stacktrace

```
Failed to retrieve user <user>
hudson.plugins.active_directory.CacheAuthenticationException: Authentication failed because there was a problem caching user <user>; nested exception is com.google.common.util.concurrent.UncheckedExecutionException: java.lang.IllegalStateException: Recursive load
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider.retrieveUser(ActiveDirectoryUnixAuthenticationProvider.java:340)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider.retrieveUser(ActiveDirectoryUnixAuthenticationProvider.java:199)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider.retrieveUser(ActiveDirectoryUnixAuthenticationProvider.java:141)
	at org.acegisecurity.providers.dao.AbstractUserDetailsAuthenticationProvider.authenticate(AbstractUserDetailsAuthenticationProvider.java:122)
	at org.acegisecurity.providers.ProviderManager.doAuthentication(ProviderManager.java:200)
	at org.acegisecurity.AbstractAuthenticationManager.authenticate(AbstractAuthenticationManager.java:47)
	at org.acegisecurity.ui.webapp.AuthenticationProcessingFilter.attemptAuthentication(AuthenticationProcessingFilter.java:74)
	at org.acegisecurity.ui.AbstractProcessingFilter.doFilter(AbstractProcessingFilter.java:252)
	at hudson.security.ChainedServletFilter$1.doFilter(ChainedServletFilter.java:87)
	at jenkins.security.BasicHeaderProcessor.doFilter(BasicHeaderProcessor.java:93)
	at hudson.security.ChainedServletFilter$1.doFilter(ChainedServletFilter.java:87)
	at org.acegisecurity.context.HttpSessionContextIntegrationFilter.doFilter(HttpSessionContextIntegrationFilter.java:249)
	at hudson.security.HttpSessionContextIntegrationFilter2.doFilter(HttpSessionContextIntegrationFilter2.java:67)
	at hudson.security.ChainedServletFilter$1.doFilter(ChainedServletFilter.java:87)
	at hudson.security.ChainedServletFilter.doFilter(ChainedServletFilter.java:76)
	at hudson.security.HudsonFilter.doFilter(HudsonFilter.java:171)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.kohsuke.stapler.compression.CompressionFilter.doFilter(CompressionFilter.java:49)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at hudson.util.CharacterEncodingFilter.doFilter(CharacterEncodingFilter.java:81)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.kohsuke.stapler.DiagnosticThreadNameFilter.doFilter(DiagnosticThreadNameFilter.java:30)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:585)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:553)
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:223)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1127)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:515)
	at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:185)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1061)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:97)
	at org.eclipse.jetty.server.Server.handle(Server.java:499)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:311)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:257)
	at org.eclipse.jetty.io.AbstractConnection$2.run(AbstractConnection.java:544)
	at winstone.BoundedExecutorService$1.run(BoundedExecutorService.java:77)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: com.google.common.util.concurrent.UncheckedExecutionException: java.lang.IllegalStateException: Recursive load
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2234)
	at com.google.common.cache.LocalCache.get(LocalCache.java:3965)
	at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:4764)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider.retrieveUser(ActiveDirectoryUnixAuthenticationProvider.java:232)
	... 40 more
Caused by: java.lang.IllegalStateException: Recursive load
	at com.google.common.base.Preconditions.checkState(Preconditions.java:145)
	at com.google.common.cache.LocalCache$Segment.waitForLoadingValue(LocalCache.java:2330)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2320)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2228)
	at com.google.common.cache.LocalCache.get(LocalCache.java:3965)
	at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:4764)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider.retrieveUser(ActiveDirectoryUnixAuthenticationProvider.java:232)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider.retrieveUser(ActiveDirectoryUnixAuthenticationProvider.java:199)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider.retrieveUser(ActiveDirectoryUnixAuthenticationProvider.java:141)
	at hudson.plugins.active_directory.AbstractActiveDirectoryAuthenticationProvider.loadUserByUsername(AbstractActiveDirectoryAuthenticationProvider.java:54)
	at hudson.plugins.active_directory.ActiveDirectorySecurityRealm.loadUserByUsername(ActiveDirectorySecurityRealm.java:731)
	at hudson.model.User$UserIDCanonicalIdResolver.resolveCanonicalId(User.java:1048)
	at hudson.model.User.get(User.java:394)
	at hudson.model.User.get(User.java:363)
	at hudson.model.User.get(User.java:483)
	at hudson.plugins.active_directory.ActiveDirectoryUserDetail.getJenkinsUser(ActiveDirectoryUserDetail.java:177)
	at hudson.plugins.active_directory.ActiveDirectoryUserDetail.updateUserInfo(ActiveDirectoryUserDetail.java:188)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider$1.call(ActiveDirectoryUnixAuthenticationProvider.java:315)
	at hudson.plugins.active_directory.ActiveDirectoryUnixAuthenticationProvider$1.call(ActiveDirectoryUnixAuthenticationProvider.java:232)
	at com.google.common.cache.LocalCache$LocalManualCache$1.load(LocalCache.java:4767)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3568)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2350)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2313)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2228)
	... 43 more
```

and it is produced because of SECURITY-243 - https://github.com/jenkinsci/jenkins/blob/jenkins-2.3/core/src/main/java/hudson/model/User.java#L1048 - Which is producing a recursive load.

@reviewbybees @andresrc @jtnord

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
@@ -229,6 +229,7 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
     public UserDetails retrieveUser(final String username, final String password, final String domainName, final List<SocketInfo> ldapServers) {
         UserDetails userDetails;
         try {
+            final ActiveDirectoryUserDetail[] cacheMiss = new ActiveDirectoryUserDetail[1];
             userDetails = userCache.get(username, new Callable<UserDetails>() {
                 public UserDetails call() throws AuthenticationException {
                     DirContext context;
@@ -312,11 +313,12 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
                         Set<GrantedAuthority> groups = resolveGroups(domainDN, dnFormatted, context);
                         groups.add(SecurityRealm.AUTHENTICATED_AUTHORITY);
 
-                        return new ActiveDirectoryUserDetail(username, password, true, true, true, true, groups.toArray(new GrantedAuthority[groups.size()]),
+                        cacheMiss[0] = new ActiveDirectoryUserDetail(username, password, true, true, true, true, groups.toArray(new GrantedAuthority[groups.size()]),
                                 getStringAttribute(user, "displayName"),
                                 getStringAttribute(user, "mail"),
                                 getStringAttribute(user, "telephoneNumber")
-                        ).updateUserInfo();
+                        );
+                        return cacheMiss[0];
                     } catch (NamingException e) {
                         if (anonymousBind && e.getMessage().contains("successful bind must be completed") && e.getMessage().contains("000004DC")) {
                             // sometimes (or always?) anonymous bind itself will succeed but the actual query will fail.
@@ -331,6 +333,9 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
                     }
                 }
             });
+            if (cacheMiss[0] != null) {
+                cacheMiss[0].updateUserInfo();
+            }
         } catch (UncheckedExecutionException e) {
            Throwable t = e.getCause();
             if (t instanceof AuthenticationException) {
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java
@@ -24,6 +24,7 @@
 package hudson.plugins.active_directory;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -177,6 +178,19 @@ public class ActiveDirectoryUserDetail extends User {
      * Gets the corresponding {@link hudson.model.User} object.
      */
     public hudson.model.User getJenkinsUser() {
+        try { // TODO 1.651.2+ remove reflection
+            return (hudson.model.User) hudson.model.User.class.getMethod("getById", String.class, boolean.class).invoke(null, getUsername(), true);
+        } catch (InvocationTargetException e) {
+            if (e.getCause() instanceof RuntimeException) {
+                throw (RuntimeException)e.getCause();
+            }
+            // Only RuntimeException is expected
+            LOGGER.log(Level.WARNING, String.format("There was a problem obtaining the Jenkins user %s by Id", getUsername()), e);
+        } catch (NoSuchMethodException e) {
+            // fine, older baseline
+        } catch (Exception e) { // unexpected
+            LOGGER.log(Level.WARNING, String.format("There was a problem obtaining the Jenkins user %s by Id", getUsername()), e);
+        }
         return hudson.model.User.get(getUsername());
     }
 


jtnord:
In [ActiveDirectoryUserDetail.updateUserInfo()](https://github.com/jenkinsci/active-directory-plugin/blob/3862baf8bcd3a88703c9ea98b7b947d9ee34d16d/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java#L180) you already know the users ID so should not call `User.get()` but [`user.getByID(String, boolean)`](http://javadoc.jenkins-ci.org/hudson/model/User.html#getById(java.lang.String, boolean))

andresrc:
@fbelzunc surely @teilo had more insight.

fbelzunc:
@jtnord If I use your suggestion then I need to have jenkins as lower version jenkins-1.651.2 in the pom.xml as your fix is very recent.

https://github.com/jenkinsci/jenkins/commit/49d10a9034b280b5e59535519c3f0d12d96c9f2d

Do you suggest to do this?

andresrc:
@fbelzunc I would keep the original fix then and include a `TODO` to update it when the baseline is incremented

fbelzunc:
@andresrc I knew someone would say this :-) @jtnord WDYT?

jtnord:
well the problem with keeping the original is IIUC that you will constantly be hitting `ActiveDirectoryUser.updateUserInfo()` which will call back into the plugin and it is only via the `User.get` self defense against recursive callbacks that you survive, so introducing a performance regression.

The plugin implementation without the latest fix is wrong, if there is any call for a TODO IMO then it should be in the pom.xml to remove the HPI hack for the lowest compatible core to be removed once the company that is behind the maintainance of this plugin no longer supports < 1.651.2.

jtnord:
when I was originally discussing implications with @jglick I recall the pom/hpi hack was what he suggested - so ping'ing him in case I remember incorrectly.

pom.xml
```diff
@@ -82,6 +82,24 @@
...
+      <plugin>
+        <groupId>org.jenkins-ci.tools</groupId>
+        <artifactId>maven-hpi-plugin</artifactId>
+        <extensions>true</extensions>
+        <configuration>
+          <archive>
+            <index>true</index>
+            <manifestEntries>
+              <Hudson-Version>1.609.18.1</Hudson-Version>
+              <Jenkins-Version>1.609.18.1</Jenkins-Version>
```
jglick: Mmm…rather dangerous. It would be much too easy to inadvertently introduce a dependency on some unrelated API added between 1.609 and 1.651.

(In https://github.com/jenkinsci/pipeline-plugin/pull/197 I used this trick only for a baseline swap _within_ an LTS line, where the risk of such an added API was quite low.)

jglick: I would rather suggest reflection for such cases:

```java
try { // TODO 1.651.2+ remove reflection
    return (hudson.model.User) hudson.model.User.class.getMethod("getById", String.class, boolean.class).invoke(getUsername(), true);
} catch (NoSuchMethodException x) {
    // fine, older baseline
} catch (Exception x) { // unexpected
    LOGGER.log(Level.WARNING, null, x);
}
return hudson.model.User.get(getUsername());
```

stephenc: +∞ on using reflection with a `//TODO to replace with direct call`

jglick:
What is the root cause here? `resolveCanonicalId` does have a (tested) defense against stack overflows. But it seems that this `SecurityRealm`’s implementation of `loadUserByUsername` is not reëntrant?

pom.xml
jtnord: problem with reflection is we must use the new call.  so it is not fine if it does not exist.
You would still need to set a baseline of jenkins to use.  so what would you use?
1.651.2 - can't plugin won;t be available for Cloudbees customers.
1.609.18.2 - can't plugin won't be buildable as we do not publish cloudbees propriatary wars...
1.609.3 - well will still blow up when the API is not available via reflection...

So you still will end up installing this plugin on something without the core update which will break the plugin. so given you need to barf we may as well just barf and keep code the same...

stephenc: If you install it on something without the core update the reflection will not find the method and fall through to the old code path, so no worse than today. On the core versions with the method, reflection will find and invoke the method and all will be well.

jtnord: >If you install it on something without the core update the reflection will not find the method and fall through to the old code path, so no worse than today.

= broken does not work....

So if you are going to accept broken "does not work" then dispense with the reflection shenanigans!

jglick: Of course one problem with a reflection-based approach is that `hpi:run` and any automated in-plugin tests will be using an old core that does not reflect the behavior of current users. Acceptance tests would be the only things catching “facepalm” regressions.

fbelzunc:
> What is the root cause here? resolveCanonicalId does have a (tested) defense against stack overflows. > But it seems that this SecurityRealm’s implementation of loadUserByUsername is not reëntrant?

@jglick right

andresrc:
@jglick Exactly, the cache is not reentrant.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java
```diff
@@ -177,6 +177,13 @@ public static long getSerialVersionUID() {
      * Gets the corresponding {@link hudson.model.User} object.
      */
     public hudson.model.User getJenkinsUser() {
+        try { // TODO 1.651.2+ remove reflection
+            return (hudson.model.User) hudson.model.User.class.getMethod("getById", String.class, boolean.class).invoke(getUsername(), true);
+        } catch (NoSuchMethodException x) {
+            // fine, older baseline
+        } catch (Exception e) { // unexpected
+            LOGGER.log(Level.WARNING, String.format("There was a problem obtaining the Jenkins user %s by Id", getUsername()), e);
+        }
```
andresrc: Maybe `InvocationTargetException` should be caught to unwind any legitimate exception `getById` may throw.

andresrc:
I would keep the `updateUserInfo` outside the `Callable` anyway to remove the risk of reentrant calls.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java
fbelzunc: @andresrc Are you sure it is necessary? `getById` should not throw any exception. So if `.invoke` fails you will just go tot the catch of `Exception` to record it. Is it not fine?

fbelzunc:
@stephenc @andresrc @jtnord This should be fine now - can I have your final review with a bee, please? ;-)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -330,7 +330,7 @@ public UserDetails call() throws AuthenticationException {
                         closeQuietly(context);
                     }
                 }
-            });
+            }).updateUserInfo();
```
stephenc: this will now cause the user info to be updated always. Previously we only updated the info if the value was not in the cache

stephenc:
I think this reintroduces a :bug: whereby the user details are continually being saved with no actual changes.  You probably want to have a `final boolean[] needsUpdate = new boolean[1];` just before [this line](https://github.com/jenkinsci/active-directory-plugin/pull/34/files#diff-14402b0dc4dd755ecdc24a8b4dee5188R232) then inside the callable you can `needsUpdate[0] = true` and finally you then do `if (needsUpdate[0]) ...`

better yet you could revert the type changes in the cache and just have a `final ActiveDirectoryUserDetail[] cacheMiss = new ActiveDirectoryUserDetail[1]` and then `if (cacheMiss[0] != null) cacheMiss[0].updateUserInfo();`

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -312,11 +317,12 @@ public UserDetails call() throws AuthenticationException {
                         Set<GrantedAuthority> groups = resolveGroups(domainDN, dnFormatted, context);
                         groups.add(SecurityRealm.AUTHENTICATED_AUTHORITY);
 
-                        return new ActiveDirectoryUserDetail(username, password, true, true, true, true, groups.toArray(new GrantedAuthority[groups.size()]),
+                        cacheMiss[0] = new ActiveDirectoryUserDetail(username, password, true, true, true, true, groups.toArray(new GrantedAuthority[groups.size()]),
```
jtnord: this does not look thread safe to me.  cacheMiss is a class level field...

jtnord:
:bug: for the threadsafety issue.  

> I would keep the updateUserInfo outside the Callable anyway to remove the risk of reentrant calls.

I would disagree :-)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -97,13 +97,18 @@
...
 
     /**
      * The {@link ActiveDirectoryGroupDetails} cache.
      */
     private final Cache<String, ActiveDirectoryGroupDetails> groupCache;
 
+    /**
+     * To determinate if we need or not to update the User Information
+     */
+    final ActiveDirectoryUserDetail[] cacheMiss = new ActiveDirectoryUserDetail[1];
```
andresrc: This must be a local variable in `retrieveUser` to avoid threading issues.

jtnord:
> I think this reintroduces a :bug: whereby the user details are continually being saved with no actual changes. 

Partially incorrect[1] - the code checks and only makes changes to the `User` object if they are different - thus no save if the `User`s details are the same. 

See - [ActiveDirectoryUserDetail.updateUserDetails()](https://github.com/jenkinsci/active-directory-plugin/blob/3862baf8bcd3a88703c9ea98b7b947d9ee34d16d/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java#L188-L203)

Now that code is not correctly checking - and if you get a cache miss due to the cache timeout etc then you are still also making saves for no valid reason as there is no check that the email is different from what is stored.  So the bug (unneeded saves) is still present - perhaps this should just be addressed.

[1] partially incorrect - as the bug is still there so not re-introduced - just hit more often :)

andresrc:
> I would disagree :-)

Normally I wouldn't agree with myself as well :), but in this specific one updates are only performed in some very specific cases, not always as @stephenc said and provides a safety net against core change behaviors like the one triggering this fix.

In any case, detecting the cache miss (correctly) is much better.

jtnord:
the whole cachemiss thing looks very threading issue prone to me if it is done outside the loop.

What if the same user is attempting to login from 2 different computers simultaneously (say automation tooling)?

Due to threading nature the User object for one could be returned before it is correctly updated if the thread that populates the cache is puased before it updates the user.
This sounds nit picking but its just a :bug: waiting to happen...

andresrc:
@jtnord nit but true. The alternative would be detecting the "reentrancy" at the `retrieveUser` method level and bypass the cache in that case. The outer cache call would provide the needed thread safety.

jtnord:
> In any case, detecting the cache miss (correctly) is much better.

It would be if you could gaurantee that a cache hit for the same user would not overtake the cache miss and population before creating the user object in `updateUserInfo`.
This is not synchronized - and I think any attempt to make it synchronized will kill performance :-o

andresrc:
> It would be if you could gaurantee that a cache hit for the same user would not overtake the cache miss and population before creating the user object in updateUserInfo.

@jtnord you are right

jtnord:
@andresrc we have not had any re-entry so far - so that bug just does not happen.
with the new API we hit the re-entry so the fix was to update the code to use the new API.

Big bugs like re-entry are easy to find and fix.  Bugs due to threading issues are a PITA to uncover.

andresrc:
> Big bugs like re-entry are easy to find and fix. Bugs due to threading issues are a PITA to uncover.

@jtnord agree

stephenc:
FFS cacheMiss should be a method scoped final variable

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
stephenc: :bug: I said add it at line 322 (now 319) not at the class level

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -97,13 +97,18 @@
     /**
      * The {@link UserDetails} cache.
      */
-    private final Cache<String, UserDetails> userCache;
+    private final Cache<String, ActiveDirectoryUserDetail> userCache;
```
stephenc: Can be a UserDetail, I suggest reverting this

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -229,8 +234,8 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
     public UserDetails retrieveUser(final String username, final String password, final String domainName, final List<SocketInfo> ldapServers) {
         UserDetails userDetails;
         try {
-            userDetails = userCache.get(username, new Callable<UserDetails>() {
-                public UserDetails call() throws AuthenticationException {
+            userDetails = userCache.get(username, new Callable<ActiveDirectoryUserDetail>() {
```
stephenc: Can be a userDetail as you will have cacheMiss (which should be defined immediately above this line) with the type info

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
stephenc: Yes I asked for a method scoped field

fbelzunc:
reflection is not done correctly - the method is static so need to start by null

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUserDetail.java
```diff
@@ -177,6 +178,19 @@ public static long getSerialVersionUID() {
      * Gets the corresponding {@link hudson.model.User} object.
      */
     public hudson.model.User getJenkinsUser() {
+        try { // TODO 1.651.2+ remove reflection
+            return (hudson.model.User) hudson.model.User.class.getMethod("getById", String.class, boolean.class).invoke(getUsername(), true);
```
stephenc: `invoke(null, getUsername(), true)`

stephenc:
:bug: in reflection

fbelzunc:
@stephenc @jtnord @andresrc I prefer not to use the `canMiss` for the reasons @jtnord explained - in any case I learned with your comments.

Now, I am just using reflection and correctly using `invoke(null, getUsername(), true)` with first argument as ´null´ as it is a static method.

Can I have your final review, please? I think now should be fine.

stephenc:
:bug: put the cacheMiss stuff back, when it is a method scoped variable there is no race condition as the closure does not escape the life of the method call. James was objecting to the use of an instance field

fbelzunc:
@stephenc I don't like a lot to have that - but I am doing it right now :-)

stephenc:
@fbelzunc you had it right and then you decided to remove for no clear reason. There are other ways to achieve the same, eg if you created the closure as a variable then the closure could have a field to hold the return value. But ultimately we need the update to be defanged from the code path it was on and the final array trick is the smallest change to affect this

fbelzunc:
@stephenc It is now update with the cacheMiss - I want to release this before it kick the ass of a lot of people :-)

stephenc:
:bee:

jtnord:
No stephen - you had it wrong.  THere is also a reace condition here as the
method is not synchronized and the same user can attempt to login from
multipe locations at the same time (e.g. automation scripts)

and only when there is a cache miss will the jenkins user be created and
updated.  however this is done after the cache has populated - so there is
a risk that the second login attempt will see the info in the cache and use
that - but the before the User has been properly initialized.  Not sure
what the issue would be with this - but it is a possibility and I would
rather debug a reentry issue which will be obvious in the stack trace than
some very subtle threadin issue which will have no obvious cause.

jtnord:
:bug:  as there is a race condition on `retreiveUser` with when the user details are set when the same user logs in from 2 different machines at the same time.

jtnord:
@stephenc @fbelzunc @andresrc perhaps we want a hangout to discuss this (and anyone else that is interested!)

fbelzunc:
@jtnord Happy to do it - I can ping you tomorrow morning.

stephenc:
If there re two threads they will both have the same object and one of them will have had a cache miss and hence do the update. Since a user "could" be updating their user record in parallel with a slow page render for that same user from a different browser, this is an exactly analogous case.

Yes it looks like a race, *but* Jenkins doesn't care about or provide a framework to prevent concurrent modification, so it should not be an issue

jtnord:
I don't really care about concurrent modification per-say (last wins so long as there is one) - e.g. the case where something gets an unpopulated `User` such that the build fails to send an email, (which is an issue) but more as where else the `User` will be created, so it may not be created such that you get `null` elsewhere which is definitely a bad thing as the second login attempt completes due to the cache lookup but before `updateUserInfo` which creates the `User` object..  (the core is explicit when you want to create a `User` vs retrieve an existing one - such that you could possible get `null` for `User.get()` when the user has just successfully logged in)

stephenc:
So the case you are concerned about requires:
* a multi-core server
* heavy thread contention (ie under above capacity load)
* a user trying to login in two separate browser sessions *at the same time*

And even then it requires a pre-emptive interrupt of the first thread (as it's not going to yield on releasing the lock)

Two of those are, IMHO, exceedingly difficult. The likely case for the second thread's page render is a glitch, or worst a stack trace that gets cleared on a page reload or worst case a logout/login again

The other side is a definite recursion issue that has been encountered.

If we want to "do the completely right thing" then we need to come at this from a different line of attack. You have not convinced me that the recursion (which will likely repeat for that user as they were unable to login and hence will re-trigger on subsequent attempts) is the lesser evil

jtnord:
does anyone run on a single core server these days :-)

> heavy thread contention (ie under above capacity load)

not really - you just need to have more threads than CPU cores in a runnable state and that is not so hard.  When the JVM switches to a different thread is non deterministic, so long as there is a single thread that is waiting to run any other thread could be paused...

> The likely case for the second thread's page render is a glitch

It may not be rendering a web page - it could be performing a a REST operation - a HTTP/503 response should not be retried by a tool.

> The other side is a definite recursion issue that has been encountered.

It is - but only because we actively changed core to call into the authentication realm on `user.get()` and we where calling `user.get` inside this code that core called into.  

Up to now there has not been any other recursion issues (it would have caused a spinning thread or a stack trace - and there are no reports of this as far as I am aware)

> The other side is a definite recursion issue that has been encountered.

Which has been fixed by using the new API `user.getById(id)`.

>  You have not convinced me that the recursion (which will likely repeat for that user as they were unable to login and hence will re-trigger on subsequent attempts) is the lesser evil

it's not a lesser evil - but with the fix we should have squashed the recursive call - so there should not be a possibility of recursion.

Yes someone in the future could change something again - but we can then change the code again, but right now I see the chance of recursion as zero and the chance of the threading issue as very very very tiny.  Very very very tiny is bigger than zero - hence my concern that what we are doing here is wrong. So the lesser evil to me is to need to make the code change again in the future to fix a bug that does not exist today vs accepting a bug that exists today but is extremely hard to reproduce - let alone understand from any stack trace you may get..

stephenc:
So the contract for `AbstractPasswordBasedSecurityRealm.authenticate(String,String)` does not mandate the creation of the `hudson.model.User` rather it is all about the `UserDetails`.

There are other security realms that return a `UserDetails` without doing any checks for the `hudson.model.User`. Thus it is perfectly valid for us to return a `UserDetails` when the corresponding `User` has not even been created. In fact for [some security realms](https://github.com/jenkinsci/pam-auth-plugin/blob/master/src/main/java/hudson/security/PAMSecurityRealm.java#L77) we can guarantee that the `User` would be `null` on first login...

So what difference does it make if we return the `ActiveDirectoryUserDetail` from the cache before the `ActiveDirectoryUserDetail` that was instantiated has done it's update?

The only difference is that other Jenkins code may race to instantiate the `hudson.model.User` object... but `User.getOrCreate()` and `User.getById()` should both be thread safe if written correctly... In other words, if there is a threading issue then that is a bug in core not in this plugin.

jtnord:
:bee:

fbelzunc:
@reviewbybees done

PR #35, opened 2016-05-23
[FIXED JENKINS-34968] Domain in login is not taken into account

fbelzunc:
If we write subdomain in the login, then this subdomain is not taken into account.

This is related to changes by this commit: https://github.com/jenkinsci/active-directory-plugin/commit/7c04f787b84cb7cd376ebb25d0b15ec649264c07

This is incorrect:
If people type in 'foo@bar' or 'bar\foo' or just 'foo', it should be treated as 'foo@bar' (where 'bar' represents the given domain name)

'bar' is represent a domain name which is set in the configuration, not in the login.

For example we set our Domain Name to 'bar.com'. Then we write login 'subbar\foo'. We expect to transform this to 'foo@subbar.bar.com'. But instead a 'foo@bar.com' is used.

https://issues.jenkins-ci.org/browse/JENKINS-34968

@reviewbybees @jtnord

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
@@ -451,17 +451,18 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
      * Returns the full user principal name of the form "joe@europe.contoso.com".
      * 
      * If people type in 'foo@bar' or 'bar\foo' or just 'foo', it should be treated as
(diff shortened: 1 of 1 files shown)


jtnord:
> For example we set our Domain Name to 'bar.com'. Then we write login 'subbar\foo'. We expect to transform this to 'foo@subbar.bar.com'. But instead a 'foo@bar.com' is used. 

That expectation is wrong to me...  if you are providing the NT legacy domain then `foo\BAR` then some magic should probably happen on that NT domain,.  If you provide a domain in the form  `foo@bar` then bar is the domain name and should be used - no further processing of the domain in this case should be performed.

PR #36, opened 2016-06-03
[JENKINS-35031] Not check password if it is NO_AUTHENTICATION

fbelzunc:
https://issues.jenkins-ci.org/browse/JENKINS-35031

@reviewbybees CC @jtnord @andresrc

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
@@ -43,6 +43,7 @@ import org.acegisecurity.providers.AuthenticationProvider;
 import org.acegisecurity.providers.UsernamePasswordAuthenticationToken;
 import org.acegisecurity.userdetails.UserDetails;
 import org.acegisecurity.userdetails.UsernameNotFoundException;
+import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang.StringUtils;
 
 import javax.naming.NamingEnumeration;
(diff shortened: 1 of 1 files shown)


jtnord:
And if the user is not cached and we go see if the user exists - do we store the password as NO_AUTHENTICATION - which would lock out that user from logging in whilst the cache is active?

https://github.com/fbelzunc/active-directory-plugin/blob/6efbc9eb30cfbb5e27b09320b5647c304c2d6062/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java#L316

andresrc:
Yes, an approach for that having the keys being not the user id but the authentication request, in an object that encapsulates, inside its equivalence relation, the different authentication circumstances.

These could lead to other issues, some possible duplication in values, but should be harmless, just a small loss of efficiency.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -228,9 +229,10 @@ private UserDetails retrieveUser(String username, UsernamePasswordAuthentication
     @SuppressFBWarnings(value = "ES_COMPARING_PARAMETER_STRING_WITH_EQ", justification = "Intentional instance check.")
     public UserDetails retrieveUser(final String username, final String password, final String domainName, final List<SocketInfo> ldapServers) {
         UserDetails userDetails;
+        String hashKey = DigestUtils.sha1Hex(username + "@@" + password);
```
jtnord: there is a possibility of collisions here - even if very small.  it would be better to leave the username component as-is this will be unique, and then to hash the password so that the key will be unique.

`String hashKey = username + "@@" + DigestUtils.sha1Hex(password);`

jtnord:
as a quick fix with the proposed change I think this will be mostly ok.

fbelzunc:
@jtnord Can I have your bee here now?

jtnord:
:bee:

fbelzunc:
@reviewbybees done

PR #37, opened 2016-06-21
[FIXED JENKINS-36041] Default  ldap timeouts and allow their configuration through the UI

fbelzunc:
Set-up Default  ldap timeouts and allow their configuration through the UI

This PR address:

* https://issues.jenkins-ci.org/browse/JENKINS-36041

@reviewbybees CC @jtnord @andresrc

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -522,7 +522,7 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
 
             for (SocketInfo ldapServer : ldapServers) {
                 try {
-                    LdapContext context = bind(principalName, password, ldapServer, props);
+                    LdapContext context = bind(principalName, password, ldapServer, newProps);
                     LOGGER.fine("Bound to " + ldapServer);
                     return context;
                 } catch (javax.naming.AuthenticationException e) {
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -513,9 +513,11 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
             // in a AD forest, it'd be mighty nice to be able to login as "joe"
             // as opposed to "joe@europe",
             // but the bind operation doesn't appear to allow me to do so.
-            props.put(Context.REFERRAL, "follow");
-            props.put("java.naming.ldap.attributes.binary","tokenGroups objectSid");
-            props.put("java.naming.ldap.factory.socket",TrustAllSocketFactory.class.getName());
+            Hashtable<String, String> newProps = new Hashtable<String, String>();
+            newProps.put(Context.REFERRAL, "follow");
+            newProps.put("java.naming.ldap.attributes.binary","tokenGroups objectSid");
+            newProps.put("java.naming.ldap.factory.socket",TrustAllSocketFactory.class.getName());
+            newProps.putAll(props);
             NamingException error = null;
 
             for (SocketInfo ldapServer : ldapServers) {
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -554,11 +554,7 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
          */
         @Deprecated
         public DirContext bind(String principalName, String password, List<SocketInfo> ldapServers) {
-            Hashtable<String, String> props = new Hashtable<String, String>();
-            props.put(Context.REFERRAL, "follow");
-            props.put("java.naming.ldap.attributes.binary", "tokenGroups objectSid");
-            props.put("java.naming.ldap.factory.socket", TrustAllSocketFactory.class.getName());
-            return bind(principalName, password, ldapServers, props);
+            return bind(principalName, password, ldapServers, new Hashtable<String, String>());
         }
 
         private void customizeLdapProperty(Hashtable<String, String> props, String propName) {
@@ -818,14 +814,14 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
         }
 
         public static Map<String,String> toMap(List<EnvironmentProperty> properties) {
+            final Map<String, String> result = new LinkedHashMap<String, String>();
             if (properties != null) {
-                final Map<String, String> result = new LinkedHashMap<String, String>();
                 for (EnvironmentProperty property:properties) {
                     result.put(property.getName(), property.getValue());
                 }
                 return result;
             }
-            return null;
+            return result;
         }
 
         @Extension
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -56,6 +56,7 @@ import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
@@ -158,7 +159,7 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
     /**
      *  Ldap extra properties
      */
-    protected List<EnvironmentProperty> extraEnvVars;
+    protected List<EnvironmentProperty> environmentProperties;
 
     public ActiveDirectorySecurityRealm(String domain, String site, String bindName, String bindPassword, String server) {
         this(domain, site, bindName, bindPassword, server, GroupLookupStrategy.AUTO, false);
@@ -172,16 +173,11 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
                                         String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups) {
         this(domain, site, bindName, bindPassword, server, groupLookupStrategy, removeIrrelevantGroups, domain!=null, null);
     }
-
-    public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
-                                        String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache) {
-        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups, customDomain, cache, null);
-    }
     
     @DataBoundConstructor
     // as Java signature, this binding doesn't make sense, so please don't use this constructor
     public ActiveDirectorySecurityRealm(String domain, String site, String bindName, String bindPassword, String server, GroupLookupStrategy groupLookupStrategy,
-                                        boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache, List<EnvironmentProperty> extraEnvVars) {
+                                        boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache) {
         if (customDomain!=null && !customDomain)
             domain = null;
         this.domain = fixEmpty(domain);
@@ -205,7 +201,11 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
 
         this.server = server;
         this.cache = cache;
-        this.extraEnvVars = extraEnvVars;
+    }
+
+    @DataBoundSetter
+    public void setEnvironmentProperties(List<EnvironmentProperty> environmentProperties) {
+        this.environmentProperties = environmentProperties;
     }
 
     @Restricted(NoExternalUse.class)
@@ -224,9 +224,10 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
         return cache == null ? null : cache.getTtl();
     }
 
+    // for jelly use only
     @Restricted(NoExternalUse.class)
-    public List<EnvironmentProperty> getExtraEnvVars() {
-        return extraEnvVars;
+    public List<EnvironmentProperty> getEnvironmentProperties() {
+        return environmentProperties;
     }
 
     public GroupLookupStrategy getGroupLookupStrategy() {
@@ -512,6 +513,9 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
             // in a AD forest, it'd be mighty nice to be able to login as "joe"
             // as opposed to "joe@europe",
             // but the bind operation doesn't appear to allow me to do so.
+            props.put(Context.REFERRAL, "follow");
+            props.put("java.naming.ldap.attributes.binary","tokenGroups objectSid");
+            props.put("java.naming.ldap.factory.socket",TrustAllSocketFactory.class.getName());
             NamingException error = null;
 
             for (SocketInfo ldapServer : ldapServers) {
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -171,13 +171,13 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
 
     public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
                                         String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups) {
-        this(domain, site, bindName, bindPassword, server, groupLookupStrategy, removeIrrelevantGroups, domain!=null, null);
+        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups,domain!=null,null);
     }
     
     @DataBoundConstructor
     // as Java signature, this binding doesn't make sense, so please don't use this constructor
-    public ActiveDirectorySecurityRealm(String domain, String site, String bindName, String bindPassword, String server, GroupLookupStrategy groupLookupStrategy,
-                                        boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache) {
+    public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
+                                        String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache) {
         if (customDomain!=null && !customDomain)
             domain = null;
         this.domain = fixEmpty(domain);
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -171,7 +171,7 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
 
     public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
                                         String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups) {
-        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups,domain!=null,null);
+        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups,domain!=null, null);
     }
     
     @DataBoundConstructor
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -151,6 +151,16 @@
...
 
+    /**
+     * Sets com.sun.jndi.ldap.connect.timeout
+     */
+    public final String ldapConnectTimeout;
+
+    /**
+     * Sets com.sun.jndi.ldap.read.timeout
+     */
+    public final String ldapReadTimeout;
```
jtnord: String?   these are numeric surely?

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -161,13 +171,18 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
 
     public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
                                         String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups) {
-        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups,domain!=null, null);
+        this(domain, site, bindName, bindPassword, server, groupLookupStrategy, removeIrrelevantGroups, domain!=null, null);
+    }
+
+    public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
```
jtnord: possibly better here to just use a `DataboudSetter` on the new fields?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +147,24 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
 
         this.userCache = cache.getUserCache();
         this.groupCache = cache.getGroupCache();
+        this.ldapConnectTimeout = realm.ldapConnectTimeout;
+        this.ldapReadTimeout = realm.ldapReadTimeout;
+
+        if (ldapConnectTimeout == null || ldapConnectTimeout.isEmpty()) {
+            System.setProperty("com.sun.jndi.ldap.connect.timeout", Long.toString(DEFAULT_LDAP_CONNECTION_TIMEOUT));
```
jtnord: Nope.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +147,24 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
...
+
+        } else {
+            System.setProperty("com.sun.jndi.ldap.connect.timeout", ldapConnectTimeout);
+
+        }
+
+        if (ldapReadTimeout == null || ldapReadTimeout.isEmpty()) {
+            System.setProperty("com.sun.jndi.ldap.read.timeout", Long.toString(DEFAULT_LDAP_READ_TIMEOUT));
```
jtnord: ditto...

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -105,6 +105,26 @@
...
+
+    /**
+     * Sets com.sun.jndi.ldap.read.timeout
+     */
+    private final String ldapReadTimeout;
+
+    /**
+     * Timeout if no connection after 30 seconds
+     */
+    private static final long DEFAULT_LDAP_CONNECTION_TIMEOUT = 30000;
```
jtnord: if someone has set the system property for their env then you are trashing what they have carefully set already...

jtnord:
:bug: on setting system properties, ignoring any values that may be set and JENKINS-36040 is not implemented (you have only exposed 2 of the many configuration options for the context...)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -105,6 +105,26 @@
      */
     private final Cache<String, ActiveDirectoryGroupDetails> groupCache;
 
+    /**
+     * Sets com.sun.jndi.ldap.connect.timeout
+     */
+    private final String ldapConnectTimeout;
```
jtnord: you do not need to change anything here.

fbelzunc:
@jtnord Regarding your comment

> JENKINS-36040 is not implemented (you have only exposed 2 of the many configuration options for the context...)

Are you expecting something like a textbox where you can set-up all of them? Shouldn't the rest be set-up through Java Arguments? What is the best way to do this in your opinion?

fbelzunc:
@jtnord I think this can address JENKINS-36041. Later on, using a TextArea ket KEY=Value JENKINS-36040 can be addressed for the rest of properties.

jtnord:
> @jtnord Regarding your comment

>>>JENKINS-36040 is not implemented (you have only exposed 2 of the many configuration options for the context...)

>Are you expecting something like a textbox where you can set-up all of them?

@fbelzunc Yes, something similar to the LDAP plugin

> Shouldn't the rest be set-up through Java Arguments? 

Well you can but then they have a global affect, and you may want to just affect the ActiveDirectory plugin.    For instance you may want to tweak the batch size (`java.naming.batchsize`) and if you did that via a system property anything else that used javax.naming would see that also.

> What is the best way to do this in your opinion?

I think the way the LDAP plugin did it is the way forward.  Let the user just add abirtary name/value pairs and put them into the context if set.
(if neither a "timeout" option has been set, nor a system property for said timeout, then specify a timeout (connect & read))

> I think this can address JENKINS-36041. Later on, using a TextArea ket KEY=Value JENKINS-36040 can be addressed for the rest of properties.

I think a list may be better than a text area.
As for addressing it later, you will have a harder time with backwards compatibility if you expose the timeout options through the UI today.  (you do not need to expose them through the UI, just use the system property value if set, and if not use a default value).

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -151,6 +151,26 @@
      */
     protected CacheConfiguration cache;
 
+    /**
+     * Sets com.sun.jndi.ldap.connect.timeout
+     */
+    public final Long ldapConnectTimeout;
```
jtnord: whatever happens - a big NONO on public fields.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -105,6 +108,21 @@
...
+
+    /**
+     * Sets com.sun.jndi.ldap.read.timeout
+     */
+    private final Long ldapReadTimeout;
+
+    /**
+     * Properties of the current context
+     */
+    private Hashtable<String, String> props = new Hashtable<String, String>();
```
jtnord: I think you are over complicating this...  the provider should not care what the timeouts or properties are.  Keep them with the descriptor, as they are not used outside of the descriptor.

fbelzunc:
@jtnord I think I addressed your comments - would you mind to review, please?

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -488,15 +508,10 @@ public FormValidation doValidate(@QueryParameter(fixEmpty = true) String domain,
...
          * otherwise broken, so try all the servers.
          */
-        public DirContext bind(String principalName, String password, List<SocketInfo> ldapServers) {
+        public DirContext bind(String principalName, String password, List<SocketInfo> ldapServers, Hashtable<String, String> props) {
             // in a AD forest, it'd be mighty nice to be able to login as "joe"
             // as opposed to "joe@europe",
             // but the bind operation doesn't appear to allow me to do so.
-            Hashtable<String, String> props = new Hashtable<String, String>();
-            props.put(Context.REFERRAL, "follow");
-            props.put("java.naming.ldap.attributes.binary","tokenGroups objectSid");
```
jtnord: So you still need to keep these - these are internal and not general properties to expose.
if they have been set then you could skip setting them rather than overwriting - but you need these defaults.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -488,15 +508,10 @@ public FormValidation doValidate(@QueryParameter(fixEmpty = true) String domain,
...
-            props.put("java.naming.ldap.factory.socket",TrustAllSocketFactory.class.getName());
```
jtnord: again - should not expect a user to have to set this.  not setting it will cause TLS issues.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -209,6 +224,11 @@ public Integer getTtl() {
         return cache == null ? null : cache.getTtl();
     }
 
+    @Restricted(NoExternalUse.class)
+    public List<EnvironmentProperty> getExtraEnvVars() {
```
jtnord: EnvVars is something in Jenkins - I would suggest a slightly different name to avoid any confision.

// also a comment on the @restricted // for jelly use only

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -161,13 +170,18 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
...
+                                        String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache) {
+        this(domain,site,bindName,bindPassword,server,groupLookupStrategy,removeIrrelevantGroups, customDomain, cache, null);
     }
     
     @DataBoundConstructor
     // as Java signature, this binding doesn't make sense, so please don't use this constructor
-    public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
-                                        String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache) {
+    public ActiveDirectorySecurityRealm(String domain, String site, String bindName, String bindPassword, String server, GroupLookupStrategy groupLookupStrategy,
+                                        boolean removeIrrelevantGroups, Boolean customDomain, CacheConfiguration cache, List<EnvironmentProperty> extraEnvVars) {
```
jtnord: you reall don't like @DataBoundSetter or am I missing something?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -105,6 +109,36 @@
...
+
+    /**
+     * Properties to be passed to the current LDAP context
+     */
+    private Hashtable<String, String> props = new Hashtable<String, String>();
+
+    /**
+     * Timeout if no connection after 30 seconds
+     */
+    protected final static String DEFAULT_LDAP_CONNECTION_TIMEOUT = "30000";
```
jtnord: I think these belong in the Descriptor.  Then let the descriptor set the property in bind if it is not set (if it has not been explicitly configured, to the System property value (if present) or the default (if sys prop is not present))

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -105,6 +109,36 @@
...
 
+    /**
+     *  Ldap extra properties added in Active Directory Configure Global Security UI
+     */
+    private List<ActiveDirectorySecurityRealm.EnvironmentProperty> extraEnvVars;
+
+    /**
+     * Properties to be passed to the current LDAP context
+     */
+    private Hashtable<String, String> props = new Hashtable<String, String>();
```
jtnord: why do we need this and extrEnvVars?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
jtnord: or maybe not....

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +161,26 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
 
         this.userCache = cache.getUserCache();
         this.groupCache = cache.getGroupCache();
+
+        props.put(Context.REFERRAL, "follow");
+        props.put("java.naming.ldap.attributes.binary","tokenGroups objectSid");
+        props.put("java.naming.ldap.factory.socket",TrustAllSocketFactory.class.getName());
+
+        this.extraEnvVars = realm.extraEnvVars;
```
jtnord: you don;t need to store this do you?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +161,26 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
...
+
+        if (extraEnvVars != null) {
+            Map<String, String> extraEnvVarsMap = ActiveDirectorySecurityRealm.EnvironmentProperty.toMap(extraEnvVars);
+            if (!extraEnvVarsMap.containsKey(LDAP_CONNECT_TIMEOUT)) {
+                System.getProperty(LDAP_CONNECT_TIMEOUT, DEFAULT_LDAP_CONNECTION_TIMEOUT);
```
jtnord: You need to do something with this result....

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +161,26 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
...
+            }
+            if (!extraEnvVarsMap.containsKey(LDAP_READ_TIMEOUT)) {
+                System.getProperty(LDAP_READ_TIMEOUT, DEFAULT_LDAP_READ_TIMEOUT);
```
jtnord: You need to do something with this result...

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +161,26 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
...
+            }
+            props.putAll(extraEnvVarsMap);
+        } else {
+            System.getProperty(LDAP_CONNECT_TIMEOUT, DEFAULT_LDAP_CONNECTION_TIMEOUT);
```
jtnord: huh?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +161,26 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
...
+            System.getProperty(LDAP_READ_TIMEOUT, DEFAULT_LDAP_READ_TIMEOUT);
```
jtnord: Hmmm....

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -105,6 +109,36 @@
...
+
+    /**
+     * Timeout if no connection after 30 seconds
+     */
+    protected final static String DEFAULT_LDAP_CONNECTION_TIMEOUT = "30000";
+
+    /**
+     * Timeout if no response after 60 seconds
+     */
+    protected final static String DEFAULT_LDAP_READ_TIMEOUT = "60000";
```
jtnord: All these variables can and should be private.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -527,6 +546,21 @@ public DirContext bind(String principalName, String password, List<SocketInfo> l
...
+         * Binds to the server using the specified username/password.
+         * <p>
+         * In a real deployment, often there are servers that don't respond or
+         * otherwise broken, so try all the servers.
+         */
+        @Deprecated
+        public DirContext bind(String principalName, String password, List<SocketInfo> ldapServers) {
+            Hashtable<String, String> props = new Hashtable<String, String>();
+            props.put(Context.REFERRAL, "follow");
+            props.put("java.naming.ldap.attributes.binary", "tokenGroups objectSid");
```
jtnord: you are adding these three anyway in the `bind(String principalName, String password, List<SocketInfo> ldapServers, Hashtable<String, String> props)` method.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -761,4 +795,47 @@ protected UserDetails authenticate(String username, String password) throws Auth
...
+
+        public static Map<String,String> toMap(List<EnvironmentProperty> properties) {
+            if (properties != null) {
+                final Map<String, String> result = new LinkedHashMap<String, String>();
+                for (EnvironmentProperty property:properties) {
+                    result.put(property.getName(), property.getValue());
+                }
+                return result;
+            }
+            return null;
```
jtnord: better to return an Empty Map, otherwise callers would need to check

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +156,27 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
 
         this.userCache = cache.getUserCache();
         this.groupCache = cache.getGroupCache();
+
+        String ldapConnectTimeout;
+        String ldapReadTimeout;
+        if (realm.environmentProperties != null) {
```
jtnord: if toMap was null safe then you could replace this if/else and simplify the code

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +156,27 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
 
         this.userCache = cache.getUserCache();
         this.groupCache = cache.getGroupCache();
+
+        String ldapConnectTimeout;
```
jtnord: ldapConnectTimeout/ldapReadTimeout can just be used inline.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -127,6 +156,27 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
...
+        String ldapReadTimeout;
+        if (realm.environmentProperties != null) {
+            Map<String, String> extraEnvVarsMap = ActiveDirectorySecurityRealm.EnvironmentProperty.toMap(realm.environmentProperties);
+            if (!extraEnvVarsMap.containsKey(LDAP_CONNECT_TIMEOUT)) {
```
jtnord: you can actually do all of this with 3 lines of code  (just call `putAll` with the map from the realm - and make the map return an empty map for null input).

```java
 props.put(LDAP_CONNECT_TIMEOUT, System.getProperty(LDAP_CONNECT_TIMEOUT, DEFAULT_LDAP_CONNECTION_TIMEOUT));
props.put(LDAP_READ_TIMEOUT, System.getProperty(LDAP_READ_TIMEOUT, DEFAULT_LDAP_READ_TIMEOUT);
props.putAll(ActiveDirectorySecurityRealm.EnvironmentProperty.toMap(realm.environmentProperties));

**Comment by jtnord** [comment]
:bee:

File: src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java [hunk 26]
```diff
@@ -127,6 +156,13 @@ public ActiveDirectoryUnixAuthenticationProvider(ActiveDirectorySecurityRealm re
 
         this.userCache = cache.getUserCache();
         this.groupCache = cache.getGroupCache();
+
+        Map<String, String> extraEnvVarsMap = ActiveDirectorySecurityRealm.EnvironmentProperty.toMap(realm.environmentProperties);
+        props.put(LDAP_CONNECT_TIMEOUT, System.getProperty(LDAP_CONNECT_TIMEOUT, DEFAULT_LDAP_CONNECTION_TIMEOUT));
+        props.put(LDAP_READ_TIMEOUT, System.getProperty(LDAP_READ_TIMEOUT, DEFAULT_LDAP_READ_TIMEOUT);
```
jtnord: missing closing `)`

fbelzunc:
@reviewbybees done

mas0061:
:+1:

PR #38, opened 2016-06-27
[JENKINS-36248] Provide an ultimate speed option.

jtnord:
If the AD setup only requires Security Groups then we already have these
from the token-groups attribute, so searching memberOf will just make
login take longer.

The exception to this is if you have users are members of a security group
via way of a intermediate distribution group, but for these cases you can
still choose one of the other 2 methods.

@reviewbybees 
[JENKINS-36248](https://issues.jenkins-ci.org/browse/JENKINS-36248)

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
@@ -482,44 +482,49 @@ public class ActiveDirectoryUnixAuthenticationProvider extends AbstractActiveDir
         if (userDN.contains("/")) {
             userDN = userDN.replace("/","\\/");
         }
-        LOGGER.finer("Looking up group of "+userDN);
-        Attributes id = context.getAttributes(userDN,new String[]{"tokenGroups","memberOf","CN"});
-        Attribute tga = id.get("tokenGroups");
-        if (tga==null) {// see JENKINS-11644. still trying to figure out when this happens
-            LOGGER.warning("Failed to retrieve tokenGroups for "+userDN);
-            HashSet<GrantedAuthority> r = new HashSet<GrantedAuthority>();
-            r.add(new GrantedAuthorityImpl("unable-to-retrieve-tokenGroups"));
-            return r;
-        }
-
-        // build up the query to retrieve all the groups
-        StringBuilder query = new StringBuilder("(|");
-        List<byte[]> sids = new ArrayList<byte[]>();
-
-        NamingEnumeration<?> tokenGroups = tga.getAll();
-        while (tokenGroups.hasMore()) {
-            byte[] gsid = (byte[])tokenGroups.next();
-            query.append("(objectSid={"+sids.size()+"})");
-            sids.add(gsid);
-        }
-        tokenGroups.close();
-
-        query.append(")");
-
         Set<GrantedAuthority> groups = new HashSet<GrantedAuthority>();
 
-        NamingEnumeration<SearchResult> renum = new LDAPSearchBuilder(context,domainDN).subTreeScope().returns("cn").search(query.toString(), sids.toArray());
-        parseMembers(userDN, groups, renum);
-        renum.close();
+        LOGGER.log(Level.FINER, "Looking up group of {0}", userDN);
+        Attributes id = context.getAttributes(userDN,new String[]{"tokenGroups","memberOf","CN"});
+        Attribute tga = id.get("tokenGroups");
 
+        if (tga==null) {
+            // tga will be null if you are not using a global catalogue
+            // or if the user is not actually a member of any security groups.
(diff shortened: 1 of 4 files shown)


stephenc:
:bee: AIUI

amuniz:
:bee: (even with all the code-style mess)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -482,44 +482,49 @@ private String getPrincipalName(String username, String domainName) {
...
-
-        query.append(")");
-
-        Set<GrantedAuthority> groups = new HashSet<GrantedAuthority>();
-
-        NamingEnumeration<SearchResult> renum = new LDAPSearchBuilder(context,domainDN).subTreeScope().returns("cn").search(query.toString(), sids.toArray());
-        parseMembers(userDN, groups, renum);
-        renum.close();
-
+        else {
```
amuniz: new line `else`... arg

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -482,44 +482,49 @@ private String getPrincipalName(String username, String domainName) {
...
-        // build up the query to retrieve all the groups
-        StringBuilder query = new StringBuilder("(|");
-        List<byte[]> sids = new ArrayList<byte[]>();
 
-        NamingEnumeration<?> tokenGroups = tga.getAll();
-        while (tokenGroups.hasMore()) {
-            byte[] gsid = (byte[])tokenGroups.next();
-            query.append("(objectSid={"+sids.size()+"})");
-            sids.add(gsid);
+        if (tga==null) {
```
amuniz: No spaces around `=` ... arg

jtnord: preserving the original code style...

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
stephenc: yep @jtnord is a fan of that style... personally I shall be delivering one quick "bop" on the head next time I see him for every time he has used this style... but it is yet to be seen whether he will be let leave the country that opted out of the world (ok the EU but same difference ;-) )

jtnord:
@reviewbybees done

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -482,44 +482,49 @@ private String getPrincipalName(String username, String domainName) {
...
+            // tga will be null if you are not using a global catalogue
+            // or if the user is not actually a member of any security groups.
+            LOGGER.log(Level.FINE, "Failed to retrieve tokenGroups for {0}", userDN);
```
fbelzunc: I am not sure about this message. If you read it, it looks like something is wrong, but most of the times you will get it because there are not any Security Groups. Perhaps, we should indicate that it can also be that there are not any tokenGroups?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -482,44 +482,49 @@ private String getPrincipalName(String username, String domainName) {
...
-
+        else {
+            // build up the query to retrieve all the groups
+            StringBuilder query = new StringBuilder("(|");
+            List<byte[]> sids = new ArrayList<byte[]>();
+    
+            NamingEnumeration<?> tokenGroups = tga.getAll();
+            while (tokenGroups.hasMore()) {
+                byte[] gsid = (byte[])tokenGroups.next();
+                query.append("(objectSid={"+sids.size()+"})");
```
fbelzunc: Shouldn't be here spaces after and before the `+`?

fbelzunc:
🐝 In case case the LDAPSearchBuilder works for that query you are building - let me know if you want me to merge this.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
jtnord: possibly - this was the original message.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
jtnord: yes - but this was just an indent change - I have not changed the rest of the code so that `git blame -w` works as expected.

jtnord:
@fbelzunc as plugin maintainer the choice to merge is up to you - but I say go for it :-)

PR #40, opened 2016-09-05
[JENKINS-37582] Not return null inside the cache

fbelzunc:
https://issues.jenkins-ci.org/browse/JENKINS-37582

@reviewbybees CC @andresrc @jtnord


oleg-nenashev:
Hmm... It may cause serious performance degradation.
I'm also not sure that usage of runtime <code> UserMayOrMayNotExistException</code> for groups (!) is a good idea. Maybe dependent logic does not handle it correctly

fbelzunc:
> Hmm... It may cause serious performance degradation. 

I don't think that for the moment it will cause performance degradation in case it is not already causing them. 

But yes, it looks like I need to better follow:
http://stackoverflow.com/questions/8298285/handle-null-value-using-guava-mapmaker-cachebuilder

oleg-nenashev:
Agreed. 🐝

andresrc:
Exactly, when you need to cache an equivalent to `null` you should use something like `Optional`.

fbelzunc:
@reviewbybees done

PR #41, opened 2016-09-16
[FIXED JENKINS-32033] Support for multiple domain controllers

fbelzunc:
This PR allows to use multiple domain controllers - even with multiple domains.

@jtnord @andresrc

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -518,7 +518,8 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
             newProps.put("java.naming.ldap.attributes.binary","tokenGroups objectSid");
             newProps.put("java.naming.ldap.factory.socket",TrustAllSocketFactory.class.getName());
             newProps.putAll(props);
-            NamingException error = null;
+            NamingException namingException = null;
+            javax.naming.AuthenticationException authenticationException = null;
 
             for (SocketInfo ldapServer : ldapServers) {
                 try {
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -518,7 +518,8 @@ public DirContext bind(String principalName, String password, List<SocketInfo> l
             newProps.put("java.naming.ldap.attributes.binary","tokenGroups objectSid");
             newProps.put("java.naming.ldap.factory.socket",TrustAllSocketFactory.class.getName());
             newProps.putAll(props);
-            NamingException error = null;
+            NamingException namingException = null;
+            javax.naming.AuthenticationException authenticationException = null;
```
andresrc: You can use `BadCredentialsException` here, as it seems the only subtype used.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -537,15 +538,18 @@ public DirContext bind(String principalName, String password, List<SocketInfo> l
...
                     LOGGER.log(Level.WARNING, "Failed to bind to "+ldapServer, e);
-                    error = e; // retry
+                    namingException = e; // retry
                 }
             }
-
             // if all the attempts failed
-            throw new BadCredentialsException("Either no such user '"+principalName+"' or incorrect password", error);
+            if (authenticationException !=null ) {
+                throw new BadCredentialsException("Either no such user '" + principalName + "' or incorrect password", authenticationException);
```
andresrc: You can throw the previously created one if you want.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -537,15 +538,18 @@ public DirContext bind(String principalName, String password, List<SocketInfo> l
...
+            } else {
+                throw new BadCredentialsException("Either no such user '" + principalName + "' or incorrect password", namingException);
+            }
```
andresrc: Ditto

jtnord:
I do not think that this was a good idea as implemented - it can lead to a user locking out their account with a single typos in the password.

fbelzunc:
I guess that because if you fail, you will be failling consecutively on all the domain controllers, what can make the user to be locked. Is this right?

PR #48, opened 2016-09-22
JENKINS-38444 Only re-try when domainControllers are specified

fbelzunc:
All the details are here:
https://github.com/jenkinsci/active-directory-plugin/pull/41

Basically, by doing this you might be locking an user account with a bad password specially in case where the ldapServers are discovered automatically by the plugin, as it will be trying to authenticate through all the tree getting BadCredentialsException and then you might be locking your account per hitting the max of retries.

I know that perhaps the right multiDomain implementation is to modify the UI to be able to link domains and corresponded domainControllers to make it easy later on the back end, but this is something I would like to do later on.
@reviewbybees @jtnord


fbelzunc:
@reviewbybees done

Closing this PR since I am going to submit today a PR with much better multi-domains support.

PR #52, opened 2016-10-31
[FIXED JENKINS-39065] Fail-over user to fallback when there are authentication issues

fbelzunc:
The intention of the following PR is to be able to have a fallback user in case there is a problem with the AD server.

The user can be configured through the Active Directory UI.

screen shot 2017-06-20 at 10 10 48

The password of this user is continously updated in the Jenkins Internal Database each time that he/she logins.


daniel-beck:
I'm not a fan of this approach. The local user database should really be a fallback for connectivity issues, and use a cached (hashed) password approach rather than hijacking the internal user database.

I'm pretty sure the presence of a local user record overriding e.g. 'disabled account' flags in AD will come as a surprise to everyone, and is likely also a security issue given how "magically" this appears to work.

fbelzunc:
So basically:

1. The local user database should really be a fallback for connectivity issues
2. Passwords should match on AD and Jenkins internal database
3. Restrict the users which will be used

daniel-beck:
Additionally, in 2.x, the 'admin' user exists and always has a password. Switching to AD currently shuts that off. AFAICT, this will result in that user being able to log in again despite AD.

PR #71, opened 2017-03-16
[FIXED JENKINS-42831] Enable StartTls is always TRUE in the UI

escoem:
Please @reviewbybees @fbelzunc 

Just and easy change:

<f:checkbox checked="true" />

by 

<f:checkbox default="true" />


fbelzunc:
🐝

oleg-nenashev:
How is it going to be "always true"? If the field value is false, it is going to be false., no?

fbelzunc:
@oleg-nenashev Yes, but on the UI it will always appears as true - although on disk it is fine.

pr71-example

jglick:
But I would advise writing a test demonstrating the bug (using `JenkinsRule.configRoundtrip()`) and confirming the fix.

fbelzunc:
@escoem Would you mind to write the test? I think it is easy.

PR #74, opened 2017-07-03
Add Docker integration tests

fbelzunc:
Submit this PR to get help from colleagues.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,10 +80,6 @@
       <scope>test</scope>
       <exclusions>
         <exclusion>
-          <groupId>com.google.guava</groupId>
-          <artifactId>guava</artifactId>
-        </exclusion>
-        <exclusion>
           <groupId>com.google.inject</groupId>
           <artifactId>guice</artifactId>
         </exclusion>
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,8 @@
 
   <properties>
     <jenkins.version>1.554.1</jenkins.version>
-    <java.level>7</java.level>
+    <java.level>6</java.level>
+    <java.level.test>7</java.level.test>
     <powermock.version>1.6.1</powermock.version>
   </properties>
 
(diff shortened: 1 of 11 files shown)


src/test/java/hudson/plugins/active_directory/docker/PlanetExpressTest.java
```diff
@@ -0,0 +1,63 @@
...
+public class PlanetExpressTest   {
+
+    @Rule
+    public DockerRule<PlanetExpress> docker = new DockerRule<PlanetExpress>(PlanetExpress.class);
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    @LocalData
```
jglick: I doubt you can use `@LocalData` in this context, since you need to wait for the container to be fully up and running before you try to use it. Just call `setSecurityRealm` programmatically instead.

fbelzunc:
@jglick What you said fixed the problem. However, now I faced a `NullPointerException`, which is coming from `impersonate` trigger by `j.jenkins.getSecurityRealm().loadUserByUsername("gogo");`

The AD plugin uses a groovy component to inject the ProviderManager - something similar is done with the ldap-plugin. However, the difference is that the AD plugin checks if Jenkins is running on Windows or Linux to see if it can do native authentication see https://github.com/jenkinsci/active-directory-plugin/blob/master/src/main/resources/hudson/plugins/active_directory/ActiveDirectory.groovy#L12

And the problem is that in user.getAuthorities() I am facing a NullPointerException when trying to impersonate because domains is null on this line of code https://github.com/jenkinsci/active-directory-plugin/blob/master/src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java#L224 - this seems to be happening because the ProviderManager is not set up with ActiveDirectoryUnixAuthenticationProvider.class.

It seems to be a problem inyecting the bean.

I am stuck here, any help is welcome.

PR #78, opened 2017-07-28
[FIXED JENKINS-36148] The test ignores the custom server if specified

fbelzunc:
[JENKINS-36148](https://issues.jenkins-ci.org/browse/JENKINS-36148)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 node('docker') {
-     stage('clean ws') {
-     deleteDir()
+stage('docker-pull') {
+        sh 'rm -rf docker-fixtures'
      }
      stage('checkout') {
         checkout scm
@@ -13,8 +13,5 @@ node('docker') {
      }
      stage('surefire-report') {
         junit 'target/surefire-reports/*.xml'
-     }
-     stage('clean ws') {
-     cleanWs()
      }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,6 +2,15 @@ node('docker') {
      stage('checkout') {
         checkout scm
      }
+     stage('use-fbelzunc-docker-fixtures') {
+             sh '''
+             git clone https://github.com/jenkinsci/docker-fixtures.git \
+             cd docker-fixtures \
+             git fetch origin pull/4/head:JENKINS-46673 \
+             git checkout JENKINS-46673
+             mvn clean install
+             '''
+    }
      stage('docker-pull') {
         sh 'docker pull fbelzunc/ad-build-container'
      }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@ node('docker') {
         checkout scm
      }
      stage('docker-pull') {
-        sh 'docker pull fbelzunc/ad-build-container'
+        sh 'docker pull fbelzunc/ad-build-container-with-docker-fixtures'
      }
      stage('maven') {
         sh 'docker run --add-host=samdom.example.com:127.0.0.1 -v /var/lib/docker --privileged --dns=127.0.0.1 --dns=8.8.8.8 -v $WORKSPACE:/project  fbelzunc/ad-build-container-with-docker-fixtures'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@ node('docker') {
      stage('checkout') {
         checkout scm
      }
-     stage('docker-pull') {
+     stage('docker-pull-latest-v1') {
         sh 'docker pull fbelzunc/ad-build-container-with-docker-fixtures'
      }
      stage('maven') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,4 +11,7 @@ node('docker') {
      stage('surefire-report') {
         junit 'target/surefire-reports/*.xml'
      }
+     stage('clean ws') {
+     cleanWs()
+     }
 }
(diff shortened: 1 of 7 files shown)


fbelzunc:
@jglick Notice that the three tests: `checkDomainHealth`, `validateNoDomain` and `validateDomain` are marked as `@Ignored` as for the moment I did not find a way in *jenkinsci* to add into the JVM DNS list the IP address where the docker container is running. However, at very least let me test this locally on my machine without too much effort.

jglick:
```
Could not resolve dependencies for project org.jenkins-ci.plugins:active-directory:hpi:2.7-SNAPSHOT: Could not find artifact org.jenkins-ci.test:docker-fixtures:jar:1.2-SNAPSHOT in azure (https://repo.azure.jenkins.io/public/)
```

You need to `mvn deploy` the upstream PR, and then use a _timestamped_ snapshot dependency here. That allows CI to build this PR, and ensures that you are picking up the same thing you built and not someone else’s unrelated `1.2-SNAPSHOT`.

jglick:
> I did not find a way in _jenkinsci_ to add into the JVM DNS list the IP address where the docker container is running

I not sure what “jenkinsci” means in this context, but anyway: you do not seem to be caling `ipBound(43)` so how could it work? It is not clear the upstream PR is even right; based on the code you needed to call to make `udpPort` work, it seems you would need to introduce `udpIpBound` as well.

fbelzunc:
@jglick I think I fixed everything.

I did the `clone/deploy` inside the build container where the ad-plugin builds.

https://github.com/fbelzunc/active-directory-plugin-build-container-with-docker-fixtures/blob/master/custom-dockerd-entrypoint.sh

Jenkinsfile
```diff
@@ -3,10 +3,10 @@ node('docker') {
         checkout scm
      }
      stage('docker-pull') {
-        sh 'docker pull fbelzunc/ad-build-container'
+        sh 'docker pull fbelzunc/ad-build-container-with-docker-fixtures'
```
jglick: Typically you want to put the fixture inside this plugin. See other plugins which define their own fixtures for examples.

pom.xml
```diff
@@ -76,7 +76,7 @@
     <dependency>
       <groupId>org.jenkins-ci.test</groupId>
       <artifactId>docker-fixtures</artifactId>
-      <version>1.0</version>
+      <version>1.4-SNAPSHOT-fbelzunc-udp-ports</version>
```
jglick: use timestamped snapshots please

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
```diff
@@ -98,6 +105,8 @@ public void setUp() throws Exception {
             }
             i ++;
         }
+        DOCKER_IP = d.ipBound(3268);
+        DOCKER_PORT = d.port(3268);
```
jglick: Should you not move these lines up to the top of the method and then use the variables in the `ActiveDirectoryDomain` constructor?

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
```diff
@@ -73,14 +78,16 @@
     public JenkinsRule j = new JenkinsRule();
 
     public final static String AD_DOMAIN = "samdom.example.com";
-    public final static String AD_MANAGER_DN = "CN=Administrator,CN=Users,DC=SAMDOM,DC=EXAMPLE,DC=COM";
+    public final static String AD_MANAGER_DN = "CN=Fred,CN=Users,DC=SAMDOM,DC=EXAMPLE,DC=COM";
     public final static String AD_MANAGER_DN_PASSWORD = "ia4uV1EeKait";
     public final static int MAX_RETRIES = 30;
+    public static String DOCKER_IP;
+    public static int DOCKER_PORT;
```
jglick: Why `static`? These should be instance variables (and conventionally lowercase).

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
```diff
@@ -163,7 +196,7 @@ public synchronized void publish(LogRecord record) {
         return logMessages;
     }
 
-    @DockerFixture(id = "ad-dc", ports= {389, 3268})
+    @DockerFixture(id = "ad-dc", ports= {53, 135, 138, 445, 39, 464, 389, 3268}, udpPorts = {53}, matchHostPorts = true)
```
jglick: Intentional to expose both TCP and UDP on 53?

Jenkinsfile
fbelzunc: @jglick The diagram below shows what I am trying to achieve.

As you can see:

* I am picking up a `docker` agent from jenkins-infra - which provides you access to the Docker Engine
* On this agent, I am pulling the docker image from [here](https://github.com/fbelzunc/active-directory-plugin-build-container-with-docker-fixtures). This docker image is a DIND image, which contains Maven in order to run `mvn clean verify` in the active-directory plugin. The reason why I am not using a `DockerFixture` is that I need a container where I can set-up the DNS `--dns=127.0.0.1 --dns=8.8.8.8 `.
* Maven docker-fixtures tests will run inside `active-directory-plugin-build-container-with-docker-fixtures`.

So this is the reason why I am not using `DockerFixtures`.

PR #83, opened 2017-11-17
Use a docker build container

fbelzunc:
In order to be able to create integration tests using the DNS of the AD controller, I am creating a build container and modifying the Jenkinsfile in order to use it.


jglick:
Revert changes to `Jenkinsfile`, and instead use testcontainers.org.

PR #84, opened 2018-02-05
Update Plugin POM to 3.3, use 1.625.3 as a minimum core requirement

oleg-nenashev:
A part of https://github.com/jenkinsci/configuration-as-code-plugin/issues/47

Some components use obsolete com4j, and it blows up the dependency check. I also bupbed the core requirement to 1.625.3, so now we do not need to bother about test/build environment differences. http://stats.jenkins.io/pluginversions/active-directory.html says that more than 95% of users run on newer core versions.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.7</version>
+    <version>3.3</version>
   </parent>
 
   <artifactId>active-directory</artifactId>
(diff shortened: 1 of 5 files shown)


oleg-nenashev:
@dwnusbaum My bad, should have checked before requesting the review. FindBugs should be fixed now,  FindBugs detectors somehow work differently on my machine. 

P.S: I cannot unsee some parts of the code, like usage of StringBuffer when setting authorities. If I ever get back to this plugin, will do some refactoring

oleg-nenashev:
@reviewbybees done

PR #86, opened 2018-09-17
[FIXED JENKINS-53576] - Improve JCASC compatibility

jonbrohauge:
Add getCustomDomain() to improve JCASC compatibility.

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -311,6 +311,11 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
         return environmentProperties;
     }
 
+    @Restricted(NoExternalUse.class)
+    public boolean getCustomDomain() {
+        return domains != null;
+    }
+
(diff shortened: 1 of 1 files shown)


jonbrohauge:
Ping @fbelzunc

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -311,6 +311,11 @@ public Integer getTtl() {
         return environmentProperties;
     }
 
+    @Restricted(NoExternalUse.class)
+    public boolean getCustomDomain() {
```
jvz: Wouldn't this make more sense named as `isCustomDomain()`?

jonbrohauge: Just trying to make a least intrusive change to make this plugin work with [Configuration as code](https://github.com/jenkinsci/configuration-as-code-plugin)

markltbaker:
@fbelzunc can the AD plugin get released so we can use it with Jenkins config as a code? Thanks!

fbelzunc:
@markltbaker Yes, I am doing it.

PR #89, opened 2019-01-28
[JENKINS-55813] Improve analysis of the AD attributes

Wadeck:
See [JENKINS-55813](https://issues.jenkins-ci.org/browse/JENKINS-55813).

PR #94, opened 2019-05-23
[JENKINS-57607] Fix regression in user details validity

jvz:
This reverts #89 so that it can be refiled as a draft PR and a hotfix can be released without it.

@fbelzunc


jvz:
Come to think of it, I can't exactly file a proper PR that un-reverts this without it being merged first.

PR #98, opened 2019-09-02
support for reverse proxy HTTP-Header based authentication

MichaelSp:
This adds a feature similar to what https://github.com/jenkinsci/reverse-proxy-auth-plugin already implements. I couldn't use that plugin however, because of https://issues.jenkins-ci.org/browse/JENKINS-29330 
I turned out to be much simpler to add the reverse-proxy authentication into *this* plugin than to add the recursive group lookup into the reverse-proxy plugin.

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -185,7 +185,7 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
     /**
      * If not null, we check if this header exists and use the username in the header to load the user.
      */
-    protected String userFromHTTPHeader;
+    protected String userFromHttpHeader;
 
     /**
      * Regular expression, if set, extracts the username out of the given userFromHTTPHeader.
@@ -826,13 +826,13 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
         return getAuthenticationProvider().loadGroupByGroupname(groupname);
     }
 
-    public String getUserFromHTTPHeader() {
-        return userFromHTTPHeader;
+    public String getUserFromHttpHeader() {
+        return userFromHttpHeader;
     }
 
     @DataBoundSetter
-    public void setUserFromHTTPHeader(String userFromHTTPHeader) {
-        this.userFromHTTPHeader = userFromHTTPHeader;
+    public void setUserFromHttpHeader(String userFromHttpHeader) {
+        this.userFromHttpHeader = userFromHttpHeader;
     }
 
     public String getUsernameExtractionExpression() {
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -188,6 +188,11 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
     protected String userFromHTTPHeader;
 
     /**
+     * Regular expression, if set, extracts the username out of the given userFromHTTPHeader.
+     */
+    protected String usernameExtractionExpression;
+
+    /**
      * Selects the SSL strategy to follow on the TLS connections
      *
      * <p>
@@ -828,6 +833,15 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
     @DataBoundSetter
     public void setUserFromHTTPHeader(String userFromHTTPHeader) {
         this.userFromHTTPHeader = userFromHTTPHeader;
+    }
+
+    public String getUsernameExtractionExpression() {
+        return usernameExtractionExpression;
+    }
+
+    @DataBoundSetter
+    public void setUsernameExtractionExpression(String usernameExtractionExpression) {
+        this.usernameExtractionExpression = usernameExtractionExpression;
     }
 
     /**
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -46,11 +46,7 @@ import org.acegisecurity.userdetails.UsernameNotFoundException;
 import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.DataBoundSetter;
-import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.StaplerRequest;
-import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.*;
 import org.kohsuke.stapler.interceptor.RequirePOST;
 import org.springframework.dao.DataAccessException;
 
@@ -80,11 +76,11 @@ import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import static hudson.Util.*;
+import static hudson.Util.fixEmpty;
 
 /**
  * {@link SecurityRealm} that talks to Active Directory.
- * 
+ *
  * @author Kohsuke Kawaguchi
  */
 public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityRealm {
@@ -122,7 +118,7 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
      * Active directory site (which specifies the physical concentration of the
      * servers), if any. If the value is non-null, we'll only contact servers in
      * this site.
-     * 
+     *
      * <p>
      * On Windows, I'm assuming ADSI takes care of everything automatically.
      *
@@ -183,6 +179,11 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
     protected List<EnvironmentProperty> environmentProperties;
 
     /**
+     * If not null, we check if this header exists and use the username in the header to load the user.
+     */
+    protected String userFromHTTPHeader;
+
+    /**
      * Selects the SSL strategy to follow on the TLS connections
      *
      * <p>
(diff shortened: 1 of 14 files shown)


MichaelSp:
Might be a good idea to check if `x-forwarded-proto` is set. What do you think?

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -74,30 +65,22 @@
...
-import java.io.Serializable;
-import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Hashtable;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.io.*;
+import java.util.*;
```
jvz: Could you fix your import settings not to collapse these?

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
```diff
@@ -0,0 +1,107 @@
...
+    HttpHeaderFilter(ActiveDirectorySecurityRealm activeDirectorySecurityRealm) {
+        this.activeDirectorySecurityRealm = activeDirectorySecurityRealm;
+    }
+
+    public void doFilter(ServletRequest servletRequest,
+                         ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        HttpServletRequest request = (HttpServletRequest) servletRequest;
+
+        if (SecurityContextHolder.getContext().getAuthentication() == null ||
```
jvz: Here you can just use `Jenkins.getAuthentication()`

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
```diff
@@ -0,0 +1,107 @@
...
+
+                    GrantedAuthority[] authorities = userDetails.getAuthorities();
+
+                    auth = new UsernamePasswordAuthenticationToken(userName, "", authorities);
+                } catch (UsernameNotFoundException e) {
+                    LOGGER.log(Level.FINE, "User from HTTP Header {0} not found in LDAP", userName);
+                }
+            }
+
+            SecurityContextHolder.getContext().setAuthentication(auth);
```
jvz: This should use `ACL.as()` in a try-with-resources block.

jvz:
This PR is looking decent. I think you might be better off adding the new constructor parameter in the `@DataBoundConstructor` as a setter method with `@DataBoundSetter` to avoid having to break backward compatibility with all the constructors you changed. Otherwise, make sure to not delete any old constructors and instead mark them `@Deprecated`.

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
MichaelSp: Not sure what you mean by that. Can you give an example?
Do you mean 
```
try (ACLContext _ = ACL.as(auth)) {
   chain.doFilter(request, response);
}
```

jvz: Yes, pretty much. `ACL.as()` is the Jenkins way of saving the current security context, setting it to the parameter, and then resetting it to the old security context in the finally block.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -195,6 +197,11 @@
      */
     protected List<EnvironmentProperty> environmentProperties;
 
+    /**
+     * If not null, we check if this header exists and use the username in the header to load the user.
+     */
+    protected String userFromHTTPHeader;
```
jvz: This should probably be named `userFromHttpHeader` to be consistent with the other initialism-based field/property name in this class.

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/help-userFromHTTPHeader.html
```diff
@@ -0,0 +1,8 @@
+<div>
+    Authenticates the user in Jenkins via a HTTP header field.
+    You can disable the behaviour if set this value to "".
+
+    SECURITY WARNING: Please make sure that this header cannot be provided
```
jvz: You can wrap this in `<div class="warning">` or `<strong class="warning-inline">` or similar to make it stand out better.

src/main/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealm/configAdvanced.jelly
```diff
@@ -1,6 +1,9 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
   <f:advanced>
+    <f:entry field="userFromHTTPHeader" title="${%Use HTTP-Header as username provider. Example: 'X-Forwarded-User'}">
```
jvz: This would be a gnarly property name to type in a translation. You can make a property name for the L10n string in `configAdvanced.properties` to make that string the default English translation. Feel free to move the other descriptions in this file there if you'd like, though simple L10n strings aren't that difficult to encode as properties file keys.

MichaelSp:
don't merge, yet. We've found a bug in the token based authentication.

MichaelSp:
false alert. I just stumbled across this one: https://jenkins.io/security/advisory/2019-07-17/#SECURITY-626 (new crumbIssuer behaviour)

jvz:
@Wadeck @jtnord thoughts on this?

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -364,6 +392,11 @@ public Authentication autoLogin(HttpServletRequest request, HttpServletResponse
         return new SecurityComponents( findBean(AuthenticationManager.class, context), uds, rms);
     }
 
+    @Override
+    public Filter createFilter(FilterConfig filterConfig) {
+        return new ChainedServletFilter(super.createFilter(filterConfig), new HttpHeaderFilter(this));
```
jtnord, line 397: so the filter is created regardless of if you want to use this new functionality or not?

(I don't recall if the filter only gets created when changing the realm, and not when just reconfiguring it)

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
```diff
@@ -0,0 +1,127 @@
...
+                    LOGGER.log(Level.FINE, "User from HTTP Header {0} not found in LDAP", userName);
+                }
+            }
+        }
+        chain.doFilter(request, response);
+    }
+
+    String getUserFromAuthorizationHeader(HttpServletRequest request) {
+        String authorization;
+        if ((authorization = request.getHeader("Authorization")) != null && authorization.toLowerCase().startsWith("basic ")) {
```
jtnord, line 68: `authorization.toLowerCase(Locale.ROOT)` ... otherwise you will get funky results in turkish

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
```diff
@@ -0,0 +1,127 @@
...
+                    return userFromHeader;
+            } else
+                return userFromHeader;
+        }
+        LOGGER.log(Level.FINE, "NO user header found");
+        return null;
+    }
+
+    private String getUserHeader() {
+        return activeDirectorySecurityRealm.userFromHttpHeader;
```
jtnord, line 111: if this is the emtpy string how does `HttpServletRequest` request perform?

MichaelSp:
thanks for your comment @jtnord 
I'm not quite sure how to go on from here. Since the mentioned ticket in `reverse proxy auth` is open since 5 years and 3 days and the last commit happened 3 years ago, it doesn't seem very plausible to be addressed there. 

If it's a matter of configuration, would you please nudge me in the right direction?

Did I mention? We're running this patch since almost 1 year live in production on ~120 Jenkins.

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
MichaelSp: from https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html#getHeader(java.lang.String)
>  If the request did not include a header of the specified name, this method returns null. 

This case is covered in line 95

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
MichaelSp: Yes it's created regardless.

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
jtnord: I am not talking about the header not being present - I am talking about `userFromHttpHeader` = `""`  
thus what does `HttpServletRequest.getHeader("")`  return?  

it is certainly undefined at the javadoc level and would be in violation of the HTTP spec - so an implementation could potentially throw an arbitrary `RuntimeException` here?

jtnord:
> I'm not quite sure how to go on from here. Since the mentioned ticket in reverse proxy auth is open since 5 years and 3 days 

the ticket could have been open that long because no one was interested enough to implement it there.

> and the last commit happened 3 years ago, it doesn't seem very plausible to be addressed there.

stable code :)  
pinging @oleg-nenashev  as he is the mainainer there (IIUC)

> If it's a matter of configuration, would you please nudge me in the right direction?

to retrieve all groups a user is a member of `(member:1.2.840.113556.1.4.1941:=cn=myuser,cn=users,DC=foo)` 

I don;t have my test setup anymore, but usign that you special OID you can the AD server to do the legwork.  (if bob is a member of `cheese` and `cheese` is a meber of `food` then the query will return both `cheese` and `food` when using the OID

oleg-nenashev:
I am not a maintainer here, sorry.

jvz:
@oleg-nenashev he was speaking about https://github.com/jenkinsci/reverse-proxy-auth-plugin which you are a maintainer of

src/main/java/hudson/plugins/active_directory/HttpHeaderFilter.java
MichaelSp: The running implementation of `HttpServletRequest.getHeader` is interchangeable. 
I've checked it in `jetty-9.4.30.v20200611`:
 1. https://github.com/eclipse/jetty.project/blob/jetty-9.4.30.v20200611/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java#L852
 2. https://github.com/eclipse/jetty.project/blob/jetty-9.4.30.v20200611/jetty-http/src/main/java/org/eclipse/jetty/http/HttpFields.java#L270-L279

I don't see the potential for a `RuntimeException`. Either there is a HTTP-Header that matches `""` than it is returned or there is no match. That should yield `null`. Both cases are handled.

MichaelSp:
outdated

PR #99, opened 2020-02-13
Add test for CasC compatibility

fcojfernandez:
`active-directory-plugin` is compatible with CasC. That plugin has an [integration test](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/integrations/src/test/java/io/jenkins/plugins/casc/ActiveDirectoryTest.java), but it's fixed to an old version. Although I plan to update it as well, this PR intends to include a test in this same plugin so in case any change broke the compatibility, it can be fixed immediately

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.40</version>
+    <version>3.56</version>
   </parent>
 
   <artifactId>active-directory</artifactId>
@@ -35,6 +35,7 @@
   <properties>
     <jenkins.version>2.60.3</jenkins.version>
     <java.level>8</java.level>
+    <configuration-as-code.version>1.35</configuration-as-code.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.40</version>
+    <version>3.56</version>
```
fcojfernandez: Needed to update some dependencies and match with JCasC

fcojfernandez:
@fbelzunc it seems that the [stage maven](https://github.com/jenkinsci/active-directory-plugin/blob/2120cde85c9aa2aa67386332474bac169ae542e8/Jenkinsfile#L10-L12) is failing trying to use maven-3.1. I'm seeing this warning and after that, a bunch of errors injecting maven objects
```
[WARNING] Error injecting: org.apache.maven.shared.transfer.artifact.install.internal.Maven31ArtifactInstaller

1) Error injecting: private org.eclipse.aether.spi.log.Logger org.apache.maven.repository.internal.DefaultVersionRangeResolver.logger
```

Any chance of updating the docker image `fbelzunc/ad-build-container-with-docker-fixtures` to use maven 3.5 and see if it's working? the `buildPlugin` execution and my own local execution rely on maven 3.5 and it succeeds.

pom.xml
alecharp: and this is good, to get the latest test-harness.

fcojfernandez: indeed that's one of the dependencies to be updated. `configuration-as-code:1.35` forces the latest version of test-harness. But updating also forced me to exclude some transitive dependencies of powermock which differ from the one used in configuration-as-code.

PR #109, opened 2021-04-23
[JENKINS-65444] remove usage of Guava

olamy:


rsandell (approved):

MRamonLeon (approved):

MRamonLeon:
Shall we mask someway Guava? shade or mask it

rsandell:
Probably wait for the caffeine plugin.

fbelzunc:
Hello @olamy , have you checked that the login and group resolution continues working in the native Windows way after this change?

olamy:
@fbelzunc  I guess it should work as it's really only technical changes.  but I can't really test on windows..

pom.xml
```diff
@@ -75,6 +75,11 @@
       <artifactId>mailer</artifactId>
       <version>1.5</version>
     </dependency>
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
+      <version>2.9.0</version>
+    </dependency>
```
jtnord, line 82: there is a caffeine API plugin incoming that you may want to use.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -217,7 +217,7 @@ public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
 
     public ActiveDirectorySecurityRealm(String domain, String site, String bindName,
                                         String bindPassword, String server, GroupLookupStrategy groupLookupStrategy, boolean removeIrrelevantGroups, CacheConfiguration cache) {
-        this(domain, Lists.newArrayList(new ActiveDirectoryDomain(domain, server)), site, bindName, bindPassword, server, groupLookupStrategy, removeIrrelevantGroups, domain!=null, cache, true);
+        this( domain, Arrays.asList(new ActiveDirectoryDomain(domain, server)), site, bindName, bindPassword, server, groupLookupStrategy, removeIrrelevantGroups, domain!=null, cache, true);
```
jtnord, line 220: isn't this a `singletonList` ?  `java.util.Collections.singletonList(T o)`  or does this list need to be modifiable?

olamy: no real idea if it's modified externally (I couldn't find any write access in the plugin code). But I would keep it modifiable as `Lists.newArrayList` does for backward compat and as even the field is public...

pom.xml
jtnord:
```suggestion
    <dependency>
      <groupId>io.jenkins.plugins</groupId>
      <artifactId>caffeine-api</artifactId>
      <version>2.9.0-17.v79e9099a51e3</version>
    </dependency>
```

jtnord (approved):

src/main/java/hudson/plugins/active_directory/ActiveDirectoryUnixAuthenticationProvider.java
```diff
@@ -574,17 +568,13 @@ public ActiveDirectoryGroupDetails call() throws NamingException {
                             }
                             LOGGER.log(Level.WARNING, "Exhausted all configured domains and could not authenticate against any");
                             throw new UserMayOrMayNotExistException(groupname);
-                        }
                     });
-        } catch (UncheckedExecutionException e) {
+        } catch (Exception e) {
```
Vlatombe, line 572: This is now catching `UserMayOrMayNotExistException` and wrapping it in `CacheAuthenticationException` which is a `RuntimeException`, causing HTTP 403 error instead of just being handled as a `UsernameNotFoundException` by the caller.

PR #171, opened 2023-07-17
[JENKINS-71667] Fortify MailAddressResolver

rsandell:
There has been cases in the wild where the `UserDetailsService` is not the expected one and so will not necessarily return an AD instance of the `UserDetails` and throw a `ClassCastException` which is not technically harmful but a fortification of the scenario is good for logging.

```[tasklist]
### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectoryMailAddressResolverImpl.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectoryMailAddressResolverImpl.java
@@ -29,8 +29,11 @@ import hudson.security.SecurityRealm;
 import hudson.tasks.MailAddressResolver;
 import jenkins.model.Jenkins;
 import org.acegisecurity.AcegiSecurityException;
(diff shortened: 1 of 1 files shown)


julieheard (commented):
You have two imports that are similar, should one be removed?
import java.util.logging.Level;
import static java.util.logging.Level.*;

fbelzunc:
@rsandell Would you mind creating a Jira for this, and renaming the PR with the Jira ID, please? After that, I will perform a release.

rsandell:
@fbelzunc done, thanks!

rsandell:
@fbelzunc This change shouldn't be too controversial right? I don't see any particular reason to wait for user feedback before releasing this?

PR #175, opened 2023-10-23
Remove prototype.js dependency

raul-arabaolaza:
There was some js code in the health status jelly page that was depending on the,  no longer provided by jenkins core,  prototype.js framework.

This change replaces the prototype usage with native js code that should work without issues with more modern jenkins versions.

See https://www.jenkins.io/blog/2023/05/12/removing-prototype-from-jenkins/

### Testing done

I have been unable to set a proper testing environment instead I have been using the js console and tools in chrome to do the tests.

So I Modified the existing script to include two new functions that I can call at will doing the same as the existing one (adding a new file to the status table). One uses prototype exactly like the existing code and the other uses the code changes included in this pr.

```             function display(r) {
                    r.domainHealth.forEach(function(domainHealth) {
                        var tr = document.createElement('tr');
                        var td1 = document.createElement('td');
                        var td2 = document.createElement('td');
                        var td3 = document.createElement('td');
                        var td4 = document.createElement('td');
                        var text1 = document.createTextNode(domainHealth.host + ' / ' + domainHealth.ipAddress);
                        var text2 = document.createTextNode(domainHealth.canLogin);
                        var text3 = document.createTextNode(domainHealth.pingExecutionTime + ' ms');
                        var text4 = document.createTextNode(domainHealth.loginExecutionTime  + ' ms');
                        td1.appendChild(text1);
                        td2.appendChild(text2);
                        td3.appendChild(text3);
                        td4.appendChild(text4);
                        tr.appendChild(td1);
                        tr.appendChild(td2);
                        tr.appendChild(td3);
                        tr.appendChild(td4);
                        $(demo).appendChild(tr);
                    });
                };

                function displayPrototype() {
                        var tr = document.createElement('tr');
                        var td1 = document.createElement('td');
                        var td2 = document.createElement('td');
                        var td3 = document.createElement('td');
                        var td4 = document.createElement('td');
                        var text1 = document.createTextNode("this");
                        var text2 = document.createTextNode("is using");
                        var text3 = document.createTextNode("prototype");
                        var text4 = document.createTextNode("and works");
                        td1.appendChild(text1);
                        td2.appendChild(text2);
                        td3.appendChild(text3);
                        td4.appendChild(text4);
                        tr.appendChild(td1);
                        tr.appendChild(td2);
                        tr.appendChild(td3);
                        tr.appendChild(td4);
                        $(demo).appendChild(tr);
                };

                function displayWithoutPrototype() {
                                        var tr = document.createElement('tr');
                                        var td1 = document.createElement('td');
                                        var td2 = document.createElement('td');
                                        var td3 = document.createElement('td');
                                        var td4 = document.createElement('td');
                                        var text1 = document.createTextNode("this");
                                        var text2 = document.createTextNode("is not using");
                                        var text3 = document.createTextNode("prototype");
                                        var text4 = document.createTextNode("and works");
                                        td1.appendChild(text1);
                                        td2.appendChild(text2);
                                        td3.appendChild(text3);
                                        td4.appendChild(text4);
                                        tr.appendChild(td1);
                                        tr.appendChild(td2);
                                        tr.appendChild(td3);
                                        tr.appendChild(td4);
                                        demo.appendChild(tr);
                                };
```

Then I locally built and installed the plugin with my changes in a development version of CBCI based on jenkins 2.426 and used the  js script console in chrome to call both new functions. That helped me confirmed that without prototype the adding code does not work and that my modifications work, see attached screenshot. 

Screenshot from 2023-10-23 17-16-58

@basil Requesting your review if you have 5 min as is very small and you are the prototype removal expert

### DIFF
--- a/src/main/resources/hudson/plugins/active_directory/ActiveDirectoryStatus/index.jelly
+++ b/src/main/resources/hudson/plugins/active_directory/ActiveDirectoryStatus/index.jelly
@@ -91,7 +91,7 @@
                         tr.appendChild(td2);
                         tr.appendChild(td3);
                         tr.appendChild(td4);
-                        $(demo).appendChild(tr);
+                        demo.appendChild(tr);
                     });
                 }
             </script>
--- a/src/main/resources/hudson/plugins/active_directory/ActiveDirectoryStatus/index.jelly
+++ b/src/main/resources/hudson/plugins/active_directory/ActiveDirectoryStatus/index.jelly
@@ -91,7 +91,7 @@
                         tr.appendChild(td2);
                         tr.appendChild(td3);
                         tr.appendChild(td4);
-                        demo.appendChild(tr);
+                        document.getElementById("demo").appendChild(tr);
                     });
                 }
             </script>


src/main/resources/hudson/plugins/active_directory/ActiveDirectoryStatus/index.jelly
```diff
@@ -91,7 +91,7 @@
                         tr.appendChild(td2);
                         tr.appendChild(td3);
                         tr.appendChild(td4);
-                        $(demo).appendChild(tr);
+                        demo.appendChild(tr);
```
rsandell, line 94: where is the `demo` var coming from?

basil: Looks like there is a `<table id="demo" class="bigtable pane">` below, so it seems the intention was to do something like `document.getElementById('demo')`. However this is a pre-existing bug that existed before this PR.

raul-arabaolaza: It is working (and was working before) because elements with id are automatically added to the `window` object and hence accessible as variables using their id as name https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object.

That said, there is no harm in improving the code a little now that we are here and use `document.getElementById()`. I will do the change

raul-arabaolaza: Changed in #03c944f

basil (approved):

PR #183, opened 2024-02-14
Adapt TheFlintstonesTest to IPv6 testing environments

imonteroperez:
### Highlights
- This PR adapts the execution of `TheFlintstonesTest` for IPv6 testing environments
- It makes feasible to use the overrided DNS as `dns://[0:0:0:0:0:0:0:1]:32780`

### Testing done
- Tested via PCT under IPv4 and IPv6 environments and working as expected.

```[tasklist]
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
```

### DIFF
--- a/src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
+++ b/src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
@@ -98,7 +98,11 @@ public class TheFlintstonesTest {
         // see hudson.plugins.active_directory.ActiveDirectoryDomain.createDNSLookupContext()
         // getHost returns a hostname not IPaddress...
         // use our DNS to resolve that to an IP address.
-        String DNS_URLs = "dns://"+InetAddress.getByName(docker.getHost()).getHostAddress()+":"+docker.getDNSPort();
+        String hostAddress = InetAddress.getByName(docker.getHost()).getHostAddress();
+        if (Boolean.getBoolean("java.net.preferIPv6Addresses")) {
+            hostAddress = String.format("[%s]", hostAddress);
+        } 
+        String DNS_URLs = "dns://"+hostAddress+":"+docker.getDNSPort();
         System.setProperty(DNSUtils.OVERRIDE_DNS_PROPERTY, DNS_URLs);
     }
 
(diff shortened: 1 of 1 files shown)


src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
```diff
@@ -98,7 +98,11 @@ public void overrideDNS() throws Exception {
         // see hudson.plugins.active_directory.ActiveDirectoryDomain.createDNSLookupContext()
         // getHost returns a hostname not IPaddress...
         // use our DNS to resolve that to an IP address.
-        String DNS_URLs = "dns://"+InetAddress.getByName(docker.getHost()).getHostAddress()+":"+docker.getDNSPort();
+        String hostAddress = InetAddress.getByName(docker.getHost()).getHostAddress();
+        if (Boolean.getBoolean("java.net.preferIPv6Addresses")) {
```
jtnord, line 102: I think this would break on IPv6 only hosts? 
as you are getting an InetAddress, you can already know if you have a v4 or v6 address and react accordingly and do not use the system property.

e.g.
```java
InetAddress inAddr = InetAddress.getByName(docker.getHost());
if (inAddr instanceof Inet6Address) {
 //...
 } else {
  /....
}
```
now if you want to do something based on a preference, you should be asking for **all** addresses (`getAllByName()`), and looking for the one according to the preference - but there may not be one...  But YAGNI


or just stop building URLs by hand... and embrace the APIs
`new URI("dns", null, docker.getHost()).getHostAddress(), docker.getDNSPort(), null, null, null).toASCIIString();`

jtnord (commented):
code appears to make some incorrect assumptions, and instead should switch on the type if InetAddress it obtains, or just not even bother and use the API

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
imonteroperez: Addressed in https://github.com/jenkinsci/active-directory-plugin/pull/183/commits/7690fd7ebd7056af5971ada488e892b781d5db3a

jtnord (approved):

PR #184, opened 2024-07-22
[JENKINS-73482] display warning when using ldap with non tls

pankajy-dev:
This PR adds a warning message on the configSecurity page when creating a configuration for Active Directory. 

### Submitter checklist
- [x] Link to relevant pull requests, esp. upstream and downstream changes


fcojfernandez (changes requested):
Already existing tests that are failing are (likely) because of the wrong condition in the constructor. Those existing tests should continue passing despite any code change

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmFipsAndForceLdapEnabledTest/testRequireTlsForWarnings/config.xml
```diff
@@ -0,0 +1,45 @@
...
+  <authorizationStrategy class="hudson.security.AuthorizationStrategy$Unsecured"/>
+  <securityRealm class="hudson.plugins.active_directory.ActiveDirectorySecurityRealm" plugin="active-directory@2.0">
+    <domains>
+      <hudson.plugins.active__directory.ActiveDirectoryDomain>
+        <name>example.com</name>
+      </hudson.plugins.active__directory.ActiveDirectoryDomain>
+    </domains>
+    <site>site</site>
+    <bindName>bindUser</bindName>
+    <bindPassword>Uv8/yXQyHku/Bx5Shl4TsTa4MJDBxBiJTSIqpEt2+DA=</bindPassword>
```
fcojfernandez, line 17: doesn't seem a real `Secret` (whose format would be like ${Uv8/yXQyHku/Bx5Shl4TsTa4MJDBxBiJTSIqpEt2+DA=}, so the `secrets/master.key` might not be needed. Could you please check?

pankajy-dev:
Warning message when application is in FIPS mode and TLS flags are disabled.
Screenshot 2024-07-24 at 4 56 29 PM

Exception when application is running in FIPS mode, force LDAP property is false, require TLS and startTls flags are disabled 
Screenshot 2024-07-24 at 4 56 15 PM

src/test/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealmFipsAndForceLdapEnabledTest.java
```diff
@@ -0,0 +1,64 @@
...
+        System.setProperty(LEGACY_FORCE_LDAPS_PROPERTY, "TRUE");
+    }
+
+    @LocalData
+    @Test
+    public void testStartTlsForWarnings() throws Exception {
+        ActiveDirectorySecurityRealm activeDirectorySecurityRealm = (ActiveDirectorySecurityRealm) jenkinsRule.jenkins.getSecurityRealm();
+
+        FormValidation result = activeDirectorySecurityRealm.getDescriptor().doCheckStartTls(false, false);
+        assertEquals(FormValidation.Kind.OK, result.kind);
```
fcojfernandez, line 35: In FIPS mode, if the two of them are false you shouldn't get an OK ... unless the System property is set. My first intention was to say this was wrong, that's why it's recommendable always to add a String with an explanation
```suggestion
        assertEquals("FIPS mode and legacy system property configured, so it's compliant", FormValidation.Kind.OK, result.kind);
```

src/test/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealmFipsAndForceLdapEnabledTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @Rule
+    public JenkinsRule jenkinsRule = new JenkinsRule();
+
+    @BeforeClass
+    public static void globalSetup() {
+        System.setProperty("jenkins.security.FIPS140.COMPLIANCE", "TRUE");
+        System.setProperty(LEGACY_FORCE_LDAPS_PROPERTY, "TRUE");
+    }
+
+    @LocalData
```
fcojfernandez, line 29: No need for a LocalData, in this test you can use `ExtensionList` to find the Descriptor class using this class: https://github.com/jenkinsci/active-directory-plugin/blob/11d154fbe2bcdad7e8a05bdeefbff746c4d837f1/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L481

src/test/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealmFipsAndForceLdapEnabledTest.java
```diff
@@ -0,0 +1,64 @@
...
+        assertEquals(FormValidation.Kind.OK, result.kind);
+
+        result = activeDirectorySecurityRealm.getDescriptor().doCheckStartTls(true, false);
+        assertEquals(FormValidation.Kind.OK, result.kind);
+
+        result = activeDirectorySecurityRealm.getDescriptor().doCheckStartTls(false, true);
+        assertEquals(FormValidation.Kind.OK, result.kind);
+    }
+
+    @LocalData
```
fcojfernandez, line 47: ditto

src/test/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealmForceLdapDisabledTest.java
fcojfernandez, line 1: Same comments about the FlagRule and the LocalData applies to this class

src/test/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealmForceLdapEnabledTest.java
fcojfernandez, line 1: Same comments about the FlagRule and the LocalData applies to this class

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testRequireTlsForWarnings/secrets/master.key
fcojfernandez, line 1: https://github.com/jenkinsci/active-directory-plugin/pull/184/files#r1688202809 same applies here. It should not be needed for this test

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testStartTlsForWarnings/secrets/master.key
fcojfernandez, line 1: https://github.com/jenkinsci/active-directory-plugin/pull/184/files#r1688202809 same applies here. It should not be needed for this test

src/main/resources/hudson/plugins/active_directory/Messages.properties
```diff
@@ -13,3 +13,4 @@ TlsConfiguration.TrustAllCertificates = (Insecure) Trust all Certificates
 TlsConfiguration.JdkTrustStore = JDK TrustStore
 
 TlsConfiguration.AdministrativeMonitor.DisplayName = Active Directory TLS Configuration Monitor
+TlsConfiguration.WarningMessage = Disabling TLS in FIPS mode is not allowed. Either enable StartTls or Require TLS.
```
jmdesprez, line 16: No need to mention `LEGACY_FORCE_LDAPS_PROPERTY` right? (I'm OK with that, I'm just asking for confirmation)

jmdesprez (changes requested):
I share the same review as @fcojfernandez (plus one question)
Especially about the use of `FlagRule` to avoid interactions with other tests :slightly_smiling_face:

src/main/resources/hudson/plugins/active_directory/Messages.properties
pankajy-dev: I did not mention `LEGACY_FORCE_LDAPS_PROPERTY` as the user controls only checkboxes while configuring.

fcojfernandez: LEGACY_FORCE_LDAPS_PROPERTY is, as it's named, legacy. It was the way of configuring TLS before the UI included the checks. For me, the text should only mention the UI

fcojfernandez:
@pankajy-dev I've been thinking and I've found a flow that might be problematic:

1. Start a fresh installation with FIPS mode and the legacy system property
2. `requiredTLS` and `startTLS` can be disabled
3. Save
4. So far, the configuration is compliant
5. Stop the instance
6. Start the instance again but without the system property

your instance now has a non compliant configuration in place. To prevent this situation I think we need to tweak the `readresolve` method

PereBueno (commented):
Agree with https://github.com/jenkinsci/active-directory-plugin/pull/184#issuecomment-2252186876 
We need to check readResolve

fcojfernandez:
after an offline discussion:

1. We drop the legacy system property from the condition
2. We are adding the check in the readresolve method just in case the xml config file is manually touched

pankajy-dev:
I'll fix the failed test cases, added check in read resolve but config.xml file not updated.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -498,8 +495,11 @@ public Object readResolve() throws ObjectStreamException {
             this.startTls = true;
         }
 
+        //requireTls can be  null, to avoid null-pointer exception, set flag as false.
+        boolean requireTlsFlag = requireTLS == null ? false : requireTLS;
```
pankajy-dev, line 499: This will cause a null pointer exception if not set to false.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -396,10 +396,7 @@ public ActiveDirectoryDomain getDomain(String domain) {
      *         <br>
      *         false if either the application is not in FIPS mode or any of requireTls, startTls is true.
      */
-    private static boolean isFipsNonCompliant(Boolean requireTls, Boolean startTls) {
-        // requireTls or startTls can be null, so in order to avoid null pointer exception assuming the values as FALSE.
```
pankajy-dev: This was giving spot bugs due to auto boxing and unboxing so reverted the change.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -396,10 +396,7 @@ public ActiveDirectoryDomain getDomain(String domain) {
...
-        requireTls = requireTls == null ? false : requireTls;
```
pankajy-dev: This was giving spot bugs due to auto boxing and unboxing so reverted the change.

pankajy-dev:
Ready for reiew.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
PereBueno: Why not simply invoking the method below with `Boolean.valueOf(requireTls)` ?

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -542,10 +569,20 @@ public ListBoxModel doFillGroupLookupStrategyItems() {
...
 
-        public FormValidation doCheckRequireTLS() {
+        public FormValidation doCheckRequireTLS(@QueryParameter boolean requireTLS, @QueryParameter boolean startTls) {
+            if (isFipsNonCompliant(requireTLS, startTls)) {
+                return FormValidation.error(Messages.TlsConfiguration_ErrorMessage());
+            }
             Jenkins.get().checkPermission(Jenkins.ADMINISTER);
             if (System.getProperty(ActiveDirectoryAuthenticationProvider.ADSI_FLAGS_SYSTEM_PROPERTY_NAME) != null) {
-                return FormValidation.warning("This setting is overridden by the ADSI mode system property");
+                return FormValidation.error("This setting is overridden by the ADSI mode system property");
```
PereBueno, line 578:
```suggestion
                return FormValidation.warning("This setting is overridden by the ADSI mode system property");
```
Sorry, I think I made the suggestion of this change myself, but we should keep existing behaviour before this change, FIPS errors are errors, overrides should be warnings

PereBueno (commented):
Overall LGTM, FIPS checks are the right ones

src/test/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest.java
```diff
@@ -325,13 +326,51 @@ public void testReadResolveMultipleDomainsOneDomainEndToEnd() throws Exception {
...
+        assertEquals(activeDirectorySecurityRealm.getCache().getTtl(), 1800);
         // Check tlsConfiguration
-        assertEquals(activeDirectorySecurityRealm.getDomains().get(0).getTlsConfiguration(), TlsConfiguration.JDK_TRUSTSTORE);
+        assertEquals(activeDirectorySecurityRealm.getDomains().get(0).getTlsConfiguration(),
+                     TlsConfiguration.JDK_TRUSTSTORE);
+    }
+
+    @LocalData
+    @Test
+    public void testStartTlsForWarnings() throws Exception {
```
jmdesprez, line 341: If this test is expected not to have FIPS enabled, you should have:
```java
    @ClassRule
    public static TestRule fip140Prop = FlagRule.systemProperty("jenkins.security.FIPS140.COMPLIANCE", "false");
```

jmdesprez (changes requested):
I agree with Pere's review. On top of that, my point of view is that setting the FlagRule in the non-FIPS test is mandatory (maybe I'm wrong). That's why I request a change.

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testReadResolveMultiDomainTwoDomainsOneDisplayName/secrets/master.key
fcojfernandez, line 1: if those files are not needed, please remove the `secrets` folder, don't leave them empty

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testFallBackUserDomain/config.xml
```diff
@@ -7,6 +7,8 @@
   <useSecurity>true</useSecurity>
   <authorizationStrategy class="hudson.security.AuthorizationStrategy$Unsecured"/>
   <securityRealm class="hudson.plugins.active_directory.ActiveDirectorySecurityRealm" plugin="active-directory@2.6">
+    <startTls>true</startTls>
+    <requireTLS>true</requireTLS>
```
fcojfernandez, line 11: why is this configuration needed now?

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -477,6 +457,53 @@ public void doAuthTest(StaplerRequest req, StaplerResponse rsp, @QueryParameter
         req.getView(this, "test.jelly").forward(req, rsp);
     }
 
+    public Object readResolve() throws ObjectStreamException {
```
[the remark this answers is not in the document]
pankajy-dev: IDE was enabled to reformat the code that I edit. I'll change the setting so that reformatting happens only for new methods.

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testFallBackUserDomain/config.xml
pankajy-dev: To make sure things are working fine with or without configs.

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
pankajy-dev: https://github.com/jenkinsci/active-directory-plugin/pull/184#discussion_r1696772071

Since this boolean value can be null, Boolean.valueOf will give NPE, which we will have to handle and then set the Boolean as FALSE in the catch. Therefore, I think using the ternary operator is better.

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testReadResolveMultiDomainTwoDomainsOneDisplayName/secrets/master.key
pankajy-dev: I removed the entire folder, diff is showing as if only content is removed.
There was an unused key in other test cases that I will remove.

jmdesprez (approved):
LGTM. https://github.com/jenkinsci/active-directory-plugin/pull/184/files#r1696772071 should be considered.

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testFallBackUserDomain/config.xml
fcojfernandez: Sorry but I don't understand this reasoning. This is actually even changing the initial configuration for the test. `null` will be as having `false`, while now it has `true`. If you had to add that because of your changes, please let us know the root cause, so we can understand the risks of having old configurations without those values

fcojfernandez (changes requested):
My concerns about the test resources are not addressed properly

pankajy-dev:
> LGTM. https://github.com/jenkinsci/active-directory-plugin/pull/184/files#r1696772071 should be considered.

https://github.com/jenkinsci/active-directory-plugin/pull/184#discussion_r1697880267
Boolean.valueOf throws NPE when null, as requireTls can be null, I am not using Boolean.valueOf here.

src/test/resources/hudson/plugins/active_directory/ActiveDirectorySecurityRealmTest/testFallBackUserDomain/config.xml
pankajy-dev: I'll change it back to the initial configuration, this is not having any impact on the outcome of the test results.

pankajy-dev:
> My concerns about the test resources are not addressed properly:
> 
> * Those `master.key` files you have emptied make no sense and actually they mean they are not needed. They are even invalid files that would make a real instance fail. Either you revert the change, or even better remove the `secrets` folders in the test resources, but please don't leave them empty.

I have not emptied the master.key files, I have actually deleted them. Unless I am misunderstanding the review. Maybe we can connect tomorrow.

pankajy-dev:
> My concerns about the test resources are not addressed properly:

> * Those `master.key` files you have emptied make no sense and actually they mean they are not needed. They are even invalid files that would make a real instance fail. Either you revert the change, or even better remove the `secrets` folders in the test resources, but please don't leave them empty.

I think when we do `load the diff` it appears like the file is not deleted and only the content is removed.

pankajy-dev:
> My concerns about the test resources are not addressed properly:
> 
> * [[JENKINS-73482] display warning when using ldap with non tls #184 (comment)](https://github.com/jenkinsci/active-directory-plugin/pull/184#discussion_r1698574055) -> those test resources where you are adding  values might mean a breaking change. If now without those values, which means `null`, the tests are failing, real instances without that configuration might fail as well. So if not needed, please remove them, otherwise we need to know if we have to document a breaking change and why

I'll change it back to the initial configuration, this is not having any impact on the outcome of the test results.

pankajy-dev:
```
	ActiveDirectorySecurityRealmTest/testFallBackUserDomain/config.xml
        ActiveDirectorySecurityRealmTest/testFallBackUserDomainController/config.xml
        ActiveDirectorySecurityRealmTest/testReadResolveMultipleDomainsOneDomainEndToEnd/config.xml
```
These files already had  `<startTls>true</startTls>`, so I have left it there.

fcojfernandez:
@pankajy-dev my bad! I misread github's message :pray:

fcojfernandez (approved):

pankajy-dev:
Thank You for the review 🙂

PR #188, opened 2024-07-24
Bump `git-changelist-maven-extension` to 1.8

jglick:
For unknown reasons, @dependabot is not offering this. Trunk builds broken trying to accessing 1.7, also for unknown reasons.

PR #189, opened 2024-07-24
chore(Jenkinsfile) stop using DinD

dduportal:
This PR provides an short term fix for https://github.com/jenkins-infra/helpdesk/issues/4194 to unblock #188 .

Main problem is a network issue in the DinD (Docker in Docker, e.g. nested container engines) preventing any download.

Long term solution would be https://github.com/jenkins-infra/helpdesk/issues/4194#issuecomment-2248329081 of course


jglick:
Or #190 more simply; this way of arranging tests is nonstandard and probably unnecessary.

dduportal:
Closing in favor of #190

jtnord:
> and probably unnecessar

nope it was very nescessary as the the AD plugi needs to do a DNS lookup and this needs to match what the AD server has otherwise it won;t work.  AD is not just LDAP it is LDAP plus DNS entries and lookups (at a simple level)

PR #190, opened 2024-07-24
Suppress Dockerized ITs since setup is broken

jglick:
https://github.com/jenkinsci/active-directory-plugin/pull/83#issuecomment-1314035127 finally causing issues https://github.com/jenkins-infra/helpdesk/issues/4194

PR #196, opened 2024-08-05
restore the Linux ITs and integrate into main build

jtnord:
effectively reverts #190 by moving the test to run as part of the main build.

### Testing done

ran `mvn test` locally on windows, observed that the `FlintstonesIT` test passed
checked the CI build contains reports for the ``FlintstonesIT` test (and that it passed where appropriate)

### Submitter checklist


Jenkinsfile
```diff
@@ -4,18 +4,16 @@ buildPlugin(useContainerAgent: false, configurations: [
   [platform: 'windows', jdk: 17],
 ])
 
-/* TODO rewrite tests to use Testcontainers (and select Linux VM nodes above)
```
jglick, line 7: Best to delete this whole block from `Jenkinsfile` and find another way.

src/test/java/hudson/plugins/active_directory/docker/ActiveDirectoryGenericContainer.java
```diff
@@ -25,7 +27,8 @@ public ActiveDirectoryGenericContainer() {
                 .withFileFromClasspath("named.conf.options", "hudson/plugins/active_directory/docker/TheFlintstonesTest/TheFlintstones/named.conf.options")
                 .withFileFromClasspath("sssd.conf", "hudson/plugins/active_directory/docker/TheFlintstonesTest/TheFlintstones/sssd.conf")
                 .withFileFromClasspath("supervisord.conf", "hudson/plugins/active_directory/docker/TheFlintstonesTest/TheFlintstones/supervisord.conf"));
-        setWaitStrategy(null);
+        // wait for the custom.sh script to complete successfully
+        setWaitStrategy(new LogMessageWaitStrategy().withRegEx(".*\\Qexited: custom (exit status 0; expected)\\E.*"));
```
jtnord, line 31: technically not part of the change, but moves the initialisation wait into the container starting to make failures easier to diagnose (which was needed during the development of this)

src/test/java/hudson/plugins/active_directory/docker/ActiveDirectoryGenericContainer.java
```diff
@@ -35,8 +38,8 @@ public ActiveDirectoryGenericContainer() {
     public ActiveDirectoryGenericContainer<SELF> withStaticPorts() {
         addFixedExposedPort(3268, 3268); // global catalog
         addFixedExposedPort(3269, 3269); // global catalog over tls
-        addFixedExposedPort(53, 53, InternetProtocol.TCP); // DNS over TCP
-        addFixedExposedPort(53, 53, InternetProtocol.UDP); // DNS over UDP
+        addFixedExposedPort(553, 53, InternetProtocol.TCP); // DNS over TCP
+        addFixedExposedPort(553, 53, InternetProtocol.UDP); // DNS over UDP
```
jtnord, line 42: works around systemd (resolved) starting a listening DNS server  (why!!???)
https://www.linuxuprising.com/2020/07/ubuntu-how-to-free-up-port-53-used-by.html  whilst there was an attempt to disable this in the CI infra, it would have caused issues anyway for local devs.  so use a different port and configure a non standard port in the code

src/test/java/hudson/plugins/active_directory/docker/EntoEndUserCacheLookupDisabledTest.java
```diff
@@ -59,9 +59,6 @@ public void customSingleADSetup(ActiveDirectoryDomain activeDirectoryDomain, Str
 
         ActiveDirectorySecurityRealm activeDirectorySecurityRealm = new ActiveDirectorySecurityRealm(null, domains, site, bindName, bindPassword, null, groupLookupStrategy, removeIrrelevantGroups, customDomain, cache, startTls, internalUsersDatabase, false);
         j.getInstance().setSecurityRealm(activeDirectorySecurityRealm);
-        while(!docker.getLogs().contains("custom (exit status 0; expected)")) {
-            Thread.sleep(1000);
-        }
```
jtnord, line 64: moved to the container wait

src/test/java/hudson/plugins/active_directory/docker/EntoEndUserCacheLookupEnabledTest.java
```diff
@@ -80,9 +80,6 @@ public void customSingleADSetup(ActiveDirectoryDomain activeDirectoryDomain, Str
 
         ActiveDirectorySecurityRealm activeDirectorySecurityRealm = new ActiveDirectorySecurityRealm(null, domains, site, bindName, bindPassword, null, groupLookupStrategy, removeIrrelevantGroups, customDomain, cache, startTls, internalUsersDatabase, false);
         j.getInstance().setSecurityRealm(activeDirectorySecurityRealm);
-        while(!docker.getLogs().contains("custom (exit status 0; expected)")) {
-            Thread.sleep(1000);
-        }
```
jtnord, line 85: moved to the container wait

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -25,32 +25,48 @@
...
+ * Integration tests with Docker and using samba as a DNS server
  */
 public class TheFlintstonesIT {
 
+
     @Rule(order = 0)
     public RequireDockerRule rdr = new RequireDockerRule();
 
+    // if the rule fails as port 53 is in use (on linux) see hack_systemd_resolve.sh 
+    // or https://www.linuxuprising.com/2020/07/ubuntu-how-to-free-up-port-53-used-by.html
```
jtnord, line 69: obsolete comment
```suggestion
```

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -60,6 +76,39 @@ public class TheFlintstonesIT {
...
+
+        HostResolutionRequestInterceptor.INSTANCE.install(
+                new MappedHostResolver(hostAliases),
+                new DNSClientHostResolver(hostIP, 553),
+                DefaultHostResolver.INSTANCE);
+
+        // we also need to set the JNDI default
+        // see hudson.plugins.active_directory.ActiveDirectoryDomain.createDNSLookupContext()
+        // getHost returns a hostname not IPaddress...
+        // use our DNS to resolve that to an IP address.
```
jtnord, line 102: confusing / obsolete comment
```suggestion
```

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -74,16 +123,13 @@ public void dynamicSetUp() throws Exception {
     }
 
     public void dynamicSetUp(boolean requireTLS) throws Exception {
-        dockerIp = requireTLS ? docker.getHost() : "dc1.samdom.example.com";
```
jtnord, line 77: this was just funky - it would use "localhost" as the server to connect to for TLS...

And yet the TLS cert for the samba server is not valid for "localhost" and caused a certificate validation exception.
As we now map `dc1.samdom.example.com` to a the docker host we will connect to it correctly and the cert will come from samba and as such be presented as `dc1.samdom.example.com`

Not sure how this worked before - most likely a change in the JVM or a change in the contain that went unmissed as the tests where not running.  Did not investigate further.

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -74,16 +123,13 @@ public void dynamicSetUp() throws Exception {
...
+        dockerIp = "dc1.samdom.example.com";
         dockerPort = docker.getMappedPort(requireTLS ? GLOBAL_CATALOG_TLS : GLOBAL_CATALOG_PLAIN_TEXT);
         ActiveDirectoryDomain activeDirectoryDomain = new ActiveDirectoryDomain(AD_DOMAIN, dockerIp + ":" +  dockerPort , null, AD_MANAGER_DN, AD_MANAGER_DN_PASSWORD);
         List<ActiveDirectoryDomain> domains = new ArrayList<>(1);
         domains.add(activeDirectoryDomain);
         ActiveDirectorySecurityRealm activeDirectorySecurityRealm = new ActiveDirectorySecurityRealm(null, domains, null, null, null, null, GroupLookupStrategy.RECURSIVE, false, true, null, false, null, requireTLS);
         j.getInstance().setSecurityRealm(activeDirectorySecurityRealm);
-        while(!docker.getLogs().contains("custom (exit status 0; expected)")) {
-            Thread.sleep(1000);
-        }
```
jtnord, line 86: moved to container startup

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesTest.java
```diff
@@ -117,9 +117,6 @@ public void dynamicSetUp() throws Exception {
         domains.add(activeDirectoryDomain);
         ActiveDirectorySecurityRealm activeDirectorySecurityRealm = new ActiveDirectorySecurityRealm(null, domains, null, null, null, null, GroupLookupStrategy.RECURSIVE, false, true, null, false, null, false);
         j.getInstance().setSecurityRealm(activeDirectorySecurityRealm);
-        while(!docker.getLogs().contains("custom (exit status 0; expected)")) {
-            Thread.sleep(1000);
-        }
```
jtnord, line 122: moved to container startup

jtnord:
Tests are running -> https://ci.jenkins.io/job/Plugins/job/active-directory-plugin/job/PR-196/12/testReport/hudson.plugins.active_directory.docker/TheFlintstonesIT/

pom.xml
```diff
@@ -129,6 +129,13 @@
       <version>1.19.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <!-- to manipulate DNS servers for testing -->
```
jglick, line 133: :cold_sweat:

pom.xml
```diff
@@ -165,13 +172,19 @@
...
+              <excludes>
+                <exclude>**/Windows*IT.java</exclude>
+              </excludes>
             </configuration>
           </plugin>
         </plugins>
       </build>
     </profile>
     <profile>
       <id>ITs</id>
```
jglick, line 184: How does this differ from `onlyITs`? Confused, could use a comment at least.

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -60,6 +74,37 @@ public class TheFlintstonesIT {
     @Rule
     public LoggerRule l = new LoggerRule();
 
+    @Before
+    public void overrideDNSServers() throws UnknownHostException {
+        // we need to pint the JVMs DNS resolver at the AD (samba) server
+        // for AD to work correctly it needs to be able to resolve hosts and do SRV lookups on the domain
+
+        // whilst the `getHost()` is supposed to return an IPAddress in some cases it will return "localhost"
```
jglick, line 82: Really??

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -60,6 +74,37 @@ public class TheFlintstonesIT {
...
+        // this adds the A entry for the PDC, but will leave the discovery of this to the SRV lookup
+
+        HostResolutionRequestInterceptor.INSTANCE.install(
+                new MappedHostResolver(hostAliases),
+                new DNSClientHostResolver(hostIP, 553),
+                DefaultHostResolver.INSTANCE);
+
+        // we also need to set the JNDI default
+        // see hudson.plugins.active_directory.ActiveDirectoryDomain.createDNSLookupContext()
+        System.setProperty(DNSUtils.OVERRIDE_DNS_PROPERTY, "dns://"+hostIP+":553");
```
jglick, line 99: Note: URL construction as string may not work on IPv6.

jglick (approved):
:crossed_fingers:

pom.xml
jtnord: better than telling people to add "server 127.0.0.1" to their resolve.conf (and disabling the systemd-resolver) just to run a test.

there was a way prior to java9 where you could set a system property and that would then work but it was removed without a reasonable replacement (you can now supply something like an `/etc/hosts` file but it will error if a host is not found in it, and it can not be used for SRV lookups

pom.xml
jtnord: `ITs` will run the "non special" ITs as well as regular tests, `onlyITs` will run **only** the "non special ITs".

The non obvious difference is that one (this one) changes global config for `surefire`, and the other adds a new `surefire` `execution`.

I declare things as special that need a specialised hard coded environment in which to run (ie the windows ITs), they will fail on any regular windows machine with docker, the windows ITs need a windows machine with docker that is joined to a domain that has a well known user with a well known password 😨 )

jtnord: now that they are run anyway, I guess we could change it to only have one profile and then let a user set the test with `-Dtest=xxx` or `-Dsurefire.includes` but I thought that @fbelzunc would prefer to keep the way he is used to.

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
jtnord: yup... returns "localhost" on my machine!

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
jtnord: good catch.

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
jtnord: https://github.com/testcontainers/testcontainers-java/blob/e3ad4ff55bbbbcc997782868b3dedb46231c5d59/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java#L463-L467

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
jtnord: e7b692d493e2b9dedd13249261afb363e5f4b8b1

pom.xml
jtnord: 5b890d8ae88c666abd8a2a687efd14066fe7abd3

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -96,7 +97,11 @@ public void overrideDNSServers() throws UnknownHostException {
 
         // we also need to set the JNDI default
         // see hudson.plugins.active_directory.ActiveDirectoryDomain.createDNSLookupContext()
-        System.setProperty(DNSUtils.OVERRIDE_DNS_PROPERTY, "dns://"+hostIP+":553");
+        if (hostInetAddr instanceof Inet6Address) {
+            System.setProperty(DNSUtils.OVERRIDE_DNS_PROPERTY, "dns://["+hostIP+"]:553");
+        } else {
+            System.setProperty(DNSUtils.OVERRIDE_DNS_PROPERTY, "dns://"+hostIP+":553");
+        }
```
jglick, line 104: or just
```suggestion
        System.setProperty(DNSUtils.OVERRIDE_DNS_PROPERTY, new URI("dns", null, hostIP, 553, null, null, null).toString());
```

src/test/java/hudson/plugins/active_directory/docker/TheFlintstonesIT.java
```diff
@@ -27,6 +27,7 @@
 import static org.junit.Assert.assertEquals;
 
 import java.io.IOException;
+import java.net.Inet6Address;
```
jglick, line 30:
```suggestion
import java.net.URI;
```

fcojfernandez (approved):

basil:
Could this please be released so that the latest release has functioning tests as part of its main build for use in PCT consumers?

PR #200, opened 2024-09-25
[JENKINS-73812] Minimum bind password length for FIPS compliance

nevingeorgesunny:
In the domain section of active directory
if you enter a short password , and the focus leaves from the field we get this message
image

is the user still clicks on on test domain it thows angry Jenkins error 

image

and if the user clicks on save , this same error appears
image 

if the user click on apply a popup with error shows
image

image


src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -266,6 +272,17 @@ public ListBoxModel doFillTlsConfigurationItems() {
             return model;
         }
 
+        /**
+         * Displays an error message if the provided password is less than 14 characters
+         * while in FIPS mode. This message is triggered when the bindPassword field loses focus.
+         */
+        public FormValidation doCheckBindPassword(@QueryParameter String bindPassword) {
```
PereBueno, line 279:
```suggestion
        @RequirePOST
        public FormValidation doCheckBindPassword(@QueryParameter String bindPassword) {
```

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -266,6 +272,17 @@ public ListBoxModel doFillTlsConfigurationItems() {
...
+            if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
```
PereBueno, line 280:
```suggestion
            Jenkins.get().checkPermission(Jenkins.ADMINISTER);
            if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
```

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
PereBueno, line 1: When performing the "Test Domain" validation we're just showing an angry jenkins with no further information.
We would have a quite better UX if we added the check in https://github.com/jenkinsci/active-directory-plugin/blob/master/src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java#L312 and simply return a `FormValidation.error()`

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainTest.java
```diff
@@ -0,0 +1,66 @@
...
+	@ClassRule
+	public static FlagRule<String> fipsSystemPropertyRule =
+			FlagRule.systemProperty("jenkins.security.FIPS140.COMPLIANCE", "true");
+
+
+	@Test
+	public void testPasswordTooShortInFIPSMode() {
+		// Create an instance of ActiveDirectoryDomain with a short password and assert exception
+		IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> {
+			new ActiveDirectoryDomain("example.com", "server", "site", "bindName", "short", TlsConfiguration.TRUST_ALL_CERTIFICATES);
```
fcojfernandez, line 27: Beware of using TRUST_ALL_CERTIFICATES. In FIPS mode it won't be allowed after https://github.com/jenkinsci/active-directory-plugin/pull/201 and the test will fail

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
fcojfernandez: you mean that in `doValidateTest` if the IllegalArgumentExcetpion is thrown, to catch it and return instead a `FormValidation.error()`, right? If so, I think that's a good idea

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -149,6 +150,11 @@ public ActiveDirectoryDomain(String name, String servers, String site, String bi
     @DataBoundConstructor
     public ActiveDirectoryDomain(String name, String servers, String site, String bindName, String bindPassword, TlsConfiguration tlsConfiguration) {
         this.name = name;
+        // Gives exception if Password is set lees than 14 chars long in FIPS mode.
+        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
+            throw new IllegalArgumentException(Messages.passwordTooShortFIPS());
```
olamy, line 155: maybe better 
```
 throw new hudson.model.Descriptor.FormException(Messages.passwordTooShortFIPS(), "bindPassword");
```

see https://issues.jenkins.io/browse/JENKINS-73771

olamy (changes requested):

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
fcojfernandez: No, please. Check MANAGE isntead to prevent a permissions error in case someone with MANAGE enters the page
```suggestion
            Jenkins.get().checkPermission(Jenkins.MANAGE);
            if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
```

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
PereBueno: yes

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
fcojfernandez: problem with that is that the FormException wasn't available when the first FIPS PRs were opened and then there would be a mismatch between PRs. Previous PR were throwing IllegalArgumentException

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
fcojfernandez: btw, MANAGE not supported in this plugin: https://github.com/jenkinsci/active-directory-plugin/pull/201/files#r1780770012

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
olamy: we did this here https://github.com/jenkinsci/credentials-plugin/pull/558 last week

rsandell: I was thinking of using `FormException` as well in the Maile plugin fix. But it is a checked exception and I didn't want to declare it in the constructor signature. So `IllegalArgumentException` seemed better imo.

rsandell (changes requested):
I would like to see some more integration type tests. Testing that calling a constructor fails is all well an good, but is not very future proof when the class evolves. 
Testing that the validation error shows up on the configuration page and that it fails when the actual form is submitted is better.
Also a test for a bad JCasC configuration should also be done.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
nevingeorgesunny: have added @RequirePOST

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
nevingeorgesunny: I Have tried this but the problem is , when I click on "Test Domain". it invokes 

```
@DataBoundConstructor
    public ActiveDirectoryDomain(String name, String servers, String site, String bindName, String bindPassword, TlsConfiguration tlsConfiguration) {
        this.name = name;
        // Gives exception if Password is set lees than 14 chars long in FIPS mode.
        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
            throw new IllegalArgumentException(Messages.passwordTooShortFIPS());
        }

        // Append default port if not specified
        servers = fixEmpty(servers);
        if (servers != null) {
            String[] serversArray = servers.split(",");
            for (int i = 0; i < serversArray.length; i++) {
                if (!serversArray[i].contains(":")) {
                    serversArray[i] += ":3268";
                }
            }
            servers = StringUtils.join(serversArray, ",");
        }
        this.servers = servers;
        this.site = fixEmpty(site);
        this.bindName = fixEmpty(bindName);
        this.bindPassword = Secret.fromString(fixEmpty(bindPassword));
        this.tlsConfiguration = tlsConfiguration;
    }
```

correct me if im wrong . but I have to do the FIPS check in the method anyway because This  is also called when I click on save or apply .

so in this check when i throw `throw new IllegalArgumentException(Messages.passwordTooShortFIPS());`. We get the angry jenkins exception.

If we need to handel this more gracefully , I think we need to add some custom  error handelling I believe ?

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainTest.java
nevingeorgesunny: I have updated this in the latest push

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
olamy: fair enough, I can live with that :) 
but if not doing it now this will probably never happen though.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
olamy: checking permission should be done for expensive check or checks including sending credentials (i.e possible leaks).
here the code is just checking length of a password so not sure it's really mandatory.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
fcojfernandez: it's done as part of https://github.com/jenkinsci/active-directory-plugin/pull/201/files#diff-501cc4984dc677ba56eb812268a11cedbd9390137ed9b274e34a30a06adb72d8R290-R303 so once merged, you'll get if for free

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
nevingeorgesunny: we can mark this as resolved right ?as we will not be checking permission

nevingeorgesunny:
> I would like to see some more integration type tests. Testing that calling a constructor fails is all well an good, but is not very future proof when the class evolves. Testing that the validation error shows up on the configuration page and that it fails when the actual form is submitted is better. Also a test for a bad JCasC configuration should also be done.
all the following test are run in FIPS mode

I have added these Integration test 
1. Using local data, small password is set , click on "test Domain" this will throw angry Jenkins error 
2. Using local data , small password is set , click on "Save"this will throw angry Jenkins error
3. Using local data , small password is set , click on "Apply"this will throw angry Jenkins error
4. Using local data , valid password is set , click on "Apply" this will return 200 then I update the password to small then i click on "Apply "this will throw angry Jenkins error

I have added some JCasC tests
1. Setting up a Jenkins with CasC YAML containing small password, this will throw error 
2. Setting up a Jenkins with CasC YAML containing valid password, this should work

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
PereBueno: Well, that's a security warn, I agree it's not that important here, so if code can be merged with that ok for me

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+//
+//		webClient.waitForBackgroundJavaScript(2000); // Wait for up to 5 seconds
+//
+//		String responseContent = resultPage.asNormalizedText();
+//		// Assert that the error message is present in the page content
+//		assertTrue(responseContent.contains("A problem occurred while processing the request"));
+//
+//		//Check that the password is too short message is present
+//		assertTrue(responseContent.contains(Messages.passwordTooShortFIPS()));
+//	}
```
fcojfernandez, line 148: why adding a commented test? please either fix it or remove it, but don't add a commented method

nevingeorgesunny: I keep getting the error [Test Domain] not found , but this does not happen locally , am i configuring something wrong ?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
olamy: correct it's a warning which can be safely ignored

fcojfernandez (approved):

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -266,6 +269,18 @@ public ListBoxModel doFillTlsConfigurationItems() {
...
+         * while in FIPS mode. This message is triggered when the bindPassword field loses focus.
+         */
+        @RequirePOST
+        public FormValidation doCheckBindPassword(@QueryParameter String bindPassword) {
+            if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
+                return FormValidation.error(Messages.passwordTooShortFIPS());
+            }
+            return FormValidation.ok();
+        }
+
```
rsandell, line 283: Missing a `readResolve` check as we've done for the other fixes.

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainIntegrationTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+	@ClassRule
+	public static FlagRule<String> fipsSystemPropertyRule =
+			FlagRule.systemProperty("jenkins.security.FIPS140.COMPLIANCE", "true");
+
+
+	/**
+	 * Tests the behavior of the "Save" button when a short password is configured.
+	 *
+	 * <p>For the preconfigured value, the password is "small" in the local data.
```
rsandell, line 39: So this test would break when adding the readResolve check.

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainIntegrationTest.java
```diff
@@ -0,0 +1,171 @@
...
+import org.jvnet.hudson.test.recipes.LocalData;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assume.assumeFalse;
+
+import static hudson.Functions.isWindows;
+
+public class ActiveDirectoryDomainIntegrationTest {
```
rsandell, line 24: These tests are explicitly for FiPS so it should be noted in the class name for easier future investigations.

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.jvnet.hudson.test.FlagRule;
+
+import hudson.util.FormValidation;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+public class ActiveDirectoryDomainTest {
```
rsandell, line 12: These tests are explicitly for FiPS so it should be noted in the class name for easier future investigations.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
fcojfernandez: as stated in several places, `readResolve` is not even needed as FIPS doesn't support backward compatibility to begin with

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainIntegrationTest.java
nevingeorgesunny: @fcojfernandez  had pushed some changes that fixed this.
I believe this can be marked as removed

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainFIPSEnabledIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+		HtmlForm form = configPage.getFormByName("config");
+
+		// Wait for JavaScript to finish loading the page
+		webClient.waitForBackgroundJavaScript(5000);
+
+		form.getInputByName("_.bindPassword").setValue("short");
+		if(!isWindows()) {
+			form.getSelectByName("_.tlsConfiguration").setSelectedAttribute("JDK_TRUSTSTORE", true);
+		}
+
```
fcojfernandez, line 116: Because of `assumeFalse("JENKINS-73847", isWindows());` this can be omitted.
```suggestion
```

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainFIPSEnabledIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+		JenkinsRule.WebClient webClient = jenkins.createWebClient();
+		// Navigate to the configuration page
+		HtmlPage configPage = webClient.goTo("configureSecurity");
+		HtmlForm form = configPage.getFormByName("config");
+
+		form.getInputByName("_.bindPassword").setValue("short");
+		if(!isWindows()) {
+			form.getSelectByName("_.tlsConfiguration").setSelectedAttribute("JDK_TRUSTSTORE", true);
+		}
```
fcojfernandez, line 74: For clarity, let's say the JENKINS ticket
```suggestion
		if(!isWindows()) {
		    // JENKINS-73847
		    form.getSelectByName("_.tlsConfiguration").setSelectedAttribute("JDK_TRUSTSTORE", true);
		}
```

fcojfernandez (approved):
Minor comments. Please apply the suggestions

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
rsandell: There is already one such check for tlsCert in this plugin so we should try to be consistent.

fcojfernandez: one check that I already explained it wasn't necessary and anyway you insisted to include making now needed that artificial consistency. Anyway, on my side I'm not converting this thread in a never ending discussion. The check was included yesterday and I'm resolving the conversation

rsandell (approved):

PR #201, opened 2024-09-26
[JENKINS-73828] when in FIPS mode, prevent user from choosing an insecure certificate

ampuscas:
In FIPS mode, insecure certificates should not be used for active directory configuration, so only trusted certificates are displayed.

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did


fcojfernandez (changes requested):
I'm afraid this won't work in CasC. CasC is based on Descriptors to find the configurations so a jenkins.yaml file could still find the Insecure Extension and use it to configure the instance in a non-compliant way

rsandell (changes requested):
I believe you need to add a check to `readResolve` as well in case the wrong configuration is already on disc before FiPS mode is enabled.

escoem:
Hi @ampuscas , can you please add the link to the JENKINS issue?

src/main/resources/hudson/plugins/active_directory/Messages.properties
```diff
@@ -13,4 +13,5 @@ TlsConfiguration.TrustAllCertificates = (Insecure) Trust all Certificates
 TlsConfiguration.JdkTrustStore = JDK TrustStore
 
 TlsConfiguration.AdministrativeMonitor.DisplayName = Active Directory TLS Configuration Monitor
-TlsConfiguration.ErrorMessage = Disabling TLS in FIPS mode is not allowed. Either enable StartTls or Require TLS.
\ No newline at end of file
+TlsConfiguration.ErrorMessage = Disabling TLS in FIPS mode is not allowed. Either enable StartTls or Require TLS.
+TlsConfiguration.CertificateError = Choosing an insecure certificate in FIPS mode is not allowed. Please choose a secure certificate.
```
fcojfernandez, line 17: Trusting all certificates is not an insecure certificate, but an insecure TLS configuration
```suggestion
TlsConfiguration.CertificateError = Choosing an insecure TLS configuration in FIPS mode is not allowed. Please do not trust all certificates.
```

src/main/java/hudson/plugins/active_directory/TlsConfiguration.java
```diff
@@ -7,8 +7,8 @@
  *
  */
 public enum TlsConfiguration {
-    TRUST_ALL_CERTIFICATES  (Messages._TlsConfiguration_TrustAllCertificates()),
-    JDK_TRUSTSTORE          (Messages._TlsConfiguration_JdkTrustStore())
+    JDK_TRUSTSTORE          (Messages._TlsConfiguration_JdkTrustStore()),
+    TRUST_ALL_CERTIFICATES  (Messages._TlsConfiguration_TrustAllCertificates())
```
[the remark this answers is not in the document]
ampuscas: Yes, the previous order displayed "Trust all certificates" by default and that caused problems if a user would add a new domain, and left this field untouched. There was no error message but an angry Jenkins page without any explanation.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -266,6 +272,15 @@ public ListBoxModel doFillTlsConfigurationItems() {
             return model;
         }
 
+        @RequirePOST
+        public FormValidation doCheckTlsConfiguration(@QueryParameter String tlsConfiguration) {
+            Jenkins.get().checkPermission(Jenkins.MANAGE);
```
fcojfernandez, line 277: CI is failing because MANAGE is not allowed in this plugin. There're 2 options here:
* add `<useBeta>true</useBeta>` to the properties in pom.xml
* Change Jenkins.MANAGE into Jenkins.ADMINISTER

ampuscas: yes, I have done that locally, but I forgot to push :(

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainFipsEnabledTest.java
fcojfernandez, line 1: the test is not valid. According to https://github.com/jenkinsci/jep/blob/master/jep/237/README.adoc#backwards-compatibility

> It is out of scope to support previous configuration, as it would not be possible to migrate non FIPS-140 compliant data to FIPS-140 compliant data within a FIPS-140 compliant environment. As such there is no backwards compatibility with previous versions.

having a previous configuration is not supported and the use of `LocalData` is not correct. If we wanted to test a previous configuration, then we should add the `readResolve` method and fail there, so even in that case, the test is not correct.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -148,6 +149,11 @@ public ActiveDirectoryDomain(String name, String servers, String site, String bi
 
     @DataBoundConstructor
     public ActiveDirectoryDomain(String name, String servers, String site, String bindName, String bindPassword, TlsConfiguration tlsConfiguration) {
+        // Gives exception if an insecure certificate is used in FIPS mode.
+        if (isFipsNonCompliant(tlsConfiguration.equals(TlsConfiguration.TRUST_ALL_CERTIFICATES))) {
```
fcojfernandez, line 153: this is the line that is causing the CI failures. If you receive a null value, then that's a NPE. The good practice is always to change the order of the condition
```suggestion
        if (isFipsNonCompliant(TlsConfiguration.TRUST_ALL_CERTIFICATES.equals(tlsConfiguration))) {
```

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -266,6 +272,15 @@ public ListBoxModel doFillTlsConfigurationItems() {
             return model;
         }
 
+        @RequirePOST
+        public FormValidation doCheckTlsConfiguration(@QueryParameter String tlsConfiguration) {
+            Jenkins.get().checkPermission(Jenkins.ADMINISTER);
+            if (!tlsConfiguration.isBlank() && isFipsNonCompliant(tlsConfiguration.equals(TlsConfiguration.TRUST_ALL_CERTIFICATES.name()))) {
```
fcojfernandez, line 278: potential NPE
```suggestion
            if (isFipsNonCompliant(TlsConfiguration.TRUST_ALL_CERTIFICATES.name().equals(tlsConfiguration))) {
```

fcojfernandez (changes requested):
NPE

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainFipsEnabledTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import io.jenkins.plugins.casc.ConfigurationAsCode;
+import io.jenkins.plugins.casc.ConfiguratorException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.fail;
+
+public class ActiveDirectoryDomainFipsEnabledTest {
+
```
rsandell, line 22: I would like to see some form of configRoundtrip test as well, imo just calling a constructor is not enough. Constructors change and gets deprecated over time and this would then just test dead code. Bu testing the form submission itself fails we make sure that it sticks around.

fcojfernandez (approved):
From my side the PR is fine. @ampuscas it seems to be missing a couple of comments from @rsandell before merging

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -197,6 +204,13 @@ public TlsConfiguration getTlsConfiguration() {
         return tlsConfiguration;
     }
 
+    private Object readResolve() throws ObjectStreamException {
```
fcojfernandez, line 207: Personally I don't like the method to be private. It is executed by reflection, so this means a violation of encapsulation. Also the throws is not needed
```suggestion
    protected Object readResolve() {
```

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainFipsEnabledTest.java
```diff
@@ -0,0 +1,78 @@
...
+        try {
+            ConfigurationAsCode.get().configure(Paths.get("src/test/resources/hudson/plugins/active_directory/ActiveDirectoryDomainFipsEnabledTest/configuration-as-code-secure.yaml").toString());
+        } catch (Exception e) {
+            fail("Secure TLS configuration should be allowed");
+        }
+    }
+
+    @Test
+    @LocalData
+    public void testBlowsUpOnStart() throws Throwable {
```
fcojfernandez, line 73: I guess @rsandell 's comment about the configRoundtrip means not only this test. It's about having a LocalData with a valid configuration and then call https://github.com/jenkinsci/jenkins-test-harness/blob/0e840c415bfa9fc16b380a74275666da356ef773/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L1433-L1438 (`jenkinsRule.configRoundtrip()`)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -197,6 +204,13 @@ public TlsConfiguration getTlsConfiguration() {
...
+        if (isFipsNonCompliant(tlsConfiguration != null && TlsConfiguration.TRUST_ALL_CERTIFICATES.name().equals(tlsConfiguration.name()))) {
```
rsandell, line 208: You could do the same enum `equals` check as in the constructor?

```suggestion
        if (isFipsNonCompliant(tlsConfiguration != null && TlsConfiguration.TRUST_ALL_CERTIFICATES.equals(tlsConfiguration))) {
```

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -148,6 +150,11 @@ public ActiveDirectoryDomain(String name, String servers, String site, String bi
 
     @DataBoundConstructor
     public ActiveDirectoryDomain(String name, String servers, String site, String bindName, String bindPassword, TlsConfiguration tlsConfiguration) {
+        // Gives exception if an insecure certificate is used in FIPS mode.
+        if (isFipsNonCompliant(TlsConfiguration.TRUST_ALL_CERTIFICATES.equals(tlsConfiguration))) {
```
rsandell, line 154: According to [this](https://github.com/jenkinsci/active-directory-plugin/blob/fa573e714719aeb65b19a4de1c6c09a8079f2f04/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L594-L596) code
```
        private boolean isTrustAllCertificatesEnabled(TlsConfiguration tlsConfiguration) {
                return (tlsConfiguration == null || TlsConfiguration.TRUST_ALL_CERTIFICATES.equals(tlsConfiguration));
        }
```
`null` also means `TRUST_ALL_CERTIFICATES` so that needs to be checked for also, I believe.

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainFipsEnabledTest.java
rsandell: I mean trying to `form.submit()` with an invalid configuration and see an Angry Jenkins judging you :)

rsandell (changes requested):
And the form submit test as well please.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
rsandell: You could probably just make that method package private and reuse it here and elsewhere.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
rsandell: So of course ignore this if you go with my suggestion above to reuse the other static method :)

src/test/java/hudson/plugins/active_directory/ActiveDirectoryDomainFipsEnabledTest.java
rsandell: mentioning configRoundtrip was unfortunately a bit misleading wording from me, it was before I realized that you can't really do a traditional configRoundtrip test because you can't create the bad backend config object to roundtrip on :)
You need to create one that is OK according to FiPS and then change the form value and submit.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
ampuscas: `tlsConfiguration` [has now the default value](https://github.com/jenkinsci/active-directory-plugin/pull/201/files#r1782288943) "JDK_TRUSTSTORE" and this is the value passed, I don't get any `null` value

rsandell: Plenty of ways to make a value `null`; It could be empty on disk (as we noticed when writing that test) it could be set to empty in casc, it could be empty with a POST to /config.xml etc. (the latter probably doesn't work on configureSecurity though) but I hope you get my point :)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -197,6 +204,13 @@ public TlsConfiguration getTlsConfiguration() {
         return tlsConfiguration;
     }
 
+    protected Object readResolve() {
+        if (isFipsNonCompliant(tlsConfiguration != null && TlsConfiguration.TRUST_ALL_CERTIFICATES.equals(tlsConfiguration))) {
```
rsandell, line 208:
```suggestion
        if (isFipsNonCompliant(tlsConfiguration != null || TlsConfiguration.TRUST_ALL_CERTIFICATES.equals(tlsConfiguration))) {
```
Probably still better to use the existing method for consistency.

rsandell (approved):

PR #203, opened 2024-09-30
[JENKINS-73827] Adding a login password length check on ActiveDirectorySecurityRealm.…

nevingeorgesunny:
If the password is shorted that 14 character and if FIPS mode is enabled , then a andry jenkins erro will be shown  swill be thrown. and login will be blocked 

image
image

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -61,6 +61,7 @@ import org.kohsuke.stapler.interceptor.RequirePOST;
 import org.kohsuke.stapler.verb.GET;
 import org.kohsuke.stapler.verb.POST;
 import org.springframework.dao.DataAccessException;
+import org.springframework.security.authentication.AuthenticationServiceException;
 
 import javax.naming.Context;
 import javax.naming.NamingEnumeration;
@@ -945,7 +946,8 @@ public class ActiveDirectorySecurityRealm extends AbstractPasswordBasedSecurityR
     protected UserDetails authenticate(String username, String password) throws AuthenticationException {
         // Check if the password length is less than 14 characters
         if(FIPS140.useCompliantAlgorithms() && StringUtils.length(password) < 14) {
-            throw new IllegalArgumentException(Messages.passwordTooShortFIPS());
+            LOGGER.log(Level.SEVERE, String.format(Messages.passwordTooShortFIPS()));
+            throw new AuthenticationServiceException(Messages.passwordTooShortFIPS());
         }
         UserDetails userDetails = getAuthenticationProvider().retrieveUser(username,new UsernamePasswordAuthenticationToken(username,password));
         SecurityListener.fireAuthenticated(userDetails);
--- a/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
+++ b/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
@@ -24,6 +24,7 @@
 package hudson.plugins.active_directory;
 
 import com4j.typelibs.ado20.ClassFactory;
+import io.jenkins.cli.shaded.org.apache.commons.lang.StringUtils;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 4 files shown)


src/test/java/hudson/plugins/active_directory/ActiveDirectoryLoginInFIPSModeTest.java
```diff
@@ -0,0 +1,88 @@
...
+		});
+
+		assertEquals(Messages.passwordTooShortFIPS(),exception.getMessage() );
+	}
+
+	@Test
+	public void testAuthenticateWithValidPassword() {
+		// Set FIPS to compliant algorithms
+		Mockito.mockStatic(FIPS140.class);
+		when(FIPS140.useCompliantAlgorithms()).thenReturn(true);
```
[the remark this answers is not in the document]
nevingeorgesunny: I have used    @ClassRule now , yes much simpler to use

src/test/java/hudson/plugins/active_directory/ActiveDirectoryLoginInFIPSModeTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.when;
+
+/**
+ * Author: Nevin Sunny
+ * Date: 30/09/24
```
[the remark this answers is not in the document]
nevingeorgesunny: removed

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -24,6 +24,7 @@
 package hudson.plugins.active_directory;
 
 import com4j.typelibs.ado20.ClassFactory;
+import io.jenkins.cli.shaded.org.apache.commons.lang.StringUtils;
```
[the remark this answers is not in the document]
nevingeorgesunny: i have updated it to the correct package

olamy (approved):
thanks for changes

src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java
```diff
@@ -941,6 +943,10 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
 
     @Override
     protected UserDetails authenticate(String username, String password) throws AuthenticationException {
+        // Check if the password length is less than 14 characters
+        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(password) < 14) {
```
jtnord, line 947: I am not sure we should just throw like this, as this leaks the fact that Jenkins is running in FIPS mode.
Not sure we can do better, as introducing a random sleep would show wildly different timing results between 14 and 15 characters for a user though.

daniel-beck: You could transform the password to be wrong in new line 950 and always fail with a "wrong password" message?

Why is it Jenkins's responsibility to reject a short password if the password is in another system? Isn't _that_ the problem?

jtnord: Just because Jenkins is running in FIPS compliance, it does not mean that any external systems are :)
(nor do they have to - we just have to be FIPS compliant when talking to them...  Many things in Jenkins would not work if it was not allowed to talk to non FIPS compliant systems,  e.g. UpdateCenter server is not FIPS compliant - but it is an externally accessed system..)

jtnord: > You could transform the password to be wrong in new line 950 and always fail with a "wrong password" message?

how would we know[1] that a password was always wrong.  IIUC there are no "this can not be in a password" characters in AD.

[1] we can guess by creating a gibberish password e.g. a "blank string of 3 characters"  but I can if I want setup a password like that on my AD server (even if I would be exceptionally  stupid to do so)

fcojfernandez: why isn't it enough to throw an `AuthenticationException` here? What's the problem exactly? The message?

jtnord: the exception is not the issue, its the fact that if you try and login with `user_that_does_not_exist:bob` and `user_that_does_not_exist:bobbobbobbobbob` by timing information (there is no query against the LDAP server in the first instance) you can detect that the system is in FIPS mode.  yet an unauthenticated user should not e able to know this information.

PereBueno: Documented in https://github.com/jenkinsci/jep/pull/409
I guess we can consider this conversation resolved?

PereBueno (approved):
LGTM

fcojfernandez (approved):

rsandell (approved):

PR #210, opened 2024-11-01
[JENKINS-74035] Extract inline script from `ActiveDirectorySecurityRealm/configAdvanced.jelly`

yaroslavafenkin:
https://issues.jenkins.io/browse/JENKINS-74035

### Testing done
Go to the Global Security Configuration page, select Active Directory as a security realm.
It's a bit weird that that code adds the event listener on `DOMContentLoaded`, because when changing the security realm to Active Directory that event has long been fired. So that code would only be executed when the realm is already set to Active Directory and then you're entering the Global Security Configuration page. Unsure if it's intentional or just a bug. But that behavior is preserved.

[Before the change](https://www.youtube.com/watch?v=XghQv6rED38)
[After the change](https://www.youtube.com/watch?v=JKDSA80e_Jw)

### Submitter checklist


basil (approved):
The existing behavior seems like a bug rather than intentional, but I don't think it has to be fixed as part of this ticket. We can file a separate ticket if it is important enough.

PR #222, opened 2025-01-25
Migrate from Acegi to Spring Security and EE 8 to EE 9

basil:
Migrate from deprecated compatibility layers to non-deprecated equivalents.


basil:
@fbelzunc Shall we proceed with this?

PR #241, opened 2025-04-24
Switch to Spring Security implementation of BCrypt

daniel-beck:
This PR switches the implementation of `BCrypt` used for the cache keys from jbcrypt to spring-security. I'm currently looking into removing the former library from core, and this is the only plugin using it.

Spring Security recently (Jenkins 2.502+) started rejecting passwords longer than 72 bytes, so this adds a bit of code to ensure we're not bumping into this limit.

This code is untested as the autotests did not manage to start containers (IIUC) on my system and I have no AD environment to test with. Happy to assist though if needed.

To test this, use passwords that are 71 (shorter than limit), 72 (exact limit), and 73+ bytes UTF-8 (~basic ASCII characters, or just ca. 18 emojis) long and expect the behavior to be as follows (unchanged from before), both on Jenkins before 2.502 and Jenkins 2.502+:

* 71: Password must match exactly
* 72: Password must match exactly or be longer (suffix is ignored)
* 73+: Password must match 72-byte prefix only, rest is ignored.

### DIFF
--- a/src/main/java/hudson/plugins/active_directory/CacheUtil.java
+++ b/src/main/java/hudson/plugins/active_directory/CacheUtil.java
@@ -1,9 +1,11 @@
 package hudson.plugins.active_directory;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


basil (approved):
Thanks for the PR!

PR #683, opened 2025-12-28
Update renovate to use Jenkins' parent config

NotMyFault:
The change proposed moves the existing renovate config to our [parent config](https://github.com/jenkinsci/renovate-config) to ease maintenance.

Custom configurations, if available, are still effective.


jtnord:
incorporated into #716

PR #693, opened 2026-01-19
Removes commons lang usage

alecharp:
Superseed #691.

This removes entirely any usage of commons lang.
The API plugin is still a transitive dependency coming from ionicons-api and configuration-as-code dependencies. Excluding it from the latter is breaking the build.

### Testing done

I ran `mvn verify` locally.

### Submitter checklist
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


mawinter69 (changes requested):

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -154,7 +151,7 @@ public ActiveDirectoryDomain(String name, String servers, String site, String bi
 
         this.name = name;
         // Gives exception if Password is set lees than 14 chars long in FIPS mode.
-        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
+        if(FIPS140.useCompliantAlgorithms() && bindPassword.length() < 14) {
```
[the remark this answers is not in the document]
alecharp: can it be null?

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -198,7 +197,7 @@ protected UserDetails retrieveUser(final String username,final  UsernamePassword
                 password = NoAuthentication.INSTANCE;
             } else {
                 final String userPassword = (String) authentication.getCredentials();
-                if (!ALLOW_EMPTY_PASSWORD && StringUtils.isEmpty(userPassword)) {
+                if (!ALLOW_EMPTY_PASSWORD && userPassword.isEmpty()) {
```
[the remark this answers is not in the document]
alecharp: > getCredentials() is annotated with nullable

`UsernamePasswordAuthenticationToken#getCredentials` from spring-security-core 6.4.4 is not and the defining interface (`Authentication`) doesn't have the annotation either. Where do you see it?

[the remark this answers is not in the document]
mawinter69: I would still add a null check here so that a BadCredentialsException is thrown which might be caught for better error handling.
And the fact that the annotation is not there in 6.x doesn't mean it can't happen. I would assume that the annotations were added to improve the automated checks.

mawinter69 (approved):

gbhat618:
just checked the build - failed test seems to be a flaky (?) build on windows has already passed.
the merge conflict is a one-liner easy one..

Perhaps @alecharp could resolve this merge conflict, there by causing a new commit - thus kicking off a new build.

If all good, maybe we can merge this PR. (I am looking to the commons-lang2 removal)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryAuthenticationProvider.java
```diff
@@ -198,7 +197,7 @@ protected UserDetails retrieveUser(final String username,final  UsernamePassword
                 password = NoAuthentication.INSTANCE;
             } else {
                 final String userPassword = (String) authentication.getCredentials();
-                if (!ALLOW_EMPTY_PASSWORD && StringUtils.isEmpty(userPassword)) {
+                if (!ALLOW_EMPTY_PASSWORD && userPassword != null && userPassword.isEmpty()) {
```
jtnord: https://javadoc.io/static/commons-lang/commons-lang/2.6/org/apache/commons/lang/StringUtils.html#isEmpty(java.lang.String)

> Checks if a String is empty ("") or null.

so this is not the same as before?

```suggestion
                if (!ALLOW_EMPTY_PASSWORD && (userPassword == null || userPassword.isEmpty())) {
```

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -350,7 +347,7 @@ public FormValidation doValidateTest(@QueryParameter(fixEmpty = true) String nam
                     return FormValidation.error("No domain was set");
                 }
 
-                if (StringUtils.isBlank(bindName)) {
+                if (bindName != null && bindName.isBlank()) {
```
mawinter69: dito
```suggestion
                if (bindName == null || bindName.isBlank()) {
```

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -154,7 +151,7 @@ public ActiveDirectoryDomain(String name, String servers, String site, String bi
 
         this.name = name;
         // Gives exception if Password is set lees than 14 chars long in FIPS mode.
-        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
+        if(FIPS140.useCompliantAlgorithms() && bindName != null && bindPassword.length() < 14) {
```
mawinter69: I think this also needs adjustment
```suggestion
        if(FIPS140.useCompliantAlgorithms() && (bindName == null || bindPassword.length() < 14)) {
```

mawinter69 (changes requested):

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
gbhat618: spotbugs failing the build here,

Redundant nullcheck of bindName, which is known to be non-null in hudson.plugins.active_directory.ActiveDirectoryDomain$DescriptorImpl.doValidateTest(String, String, String, String, String, TlsConfiguration, GroupLookupStrategy, boolean, boolean, boolean)

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -154,7 +150,7 @@ public ActiveDirectoryDomain(String name, String servers, String site, String bi
 
         this.name = name;
         // Gives exception if Password is set lees than 14 chars long in FIPS mode.
-        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
+        if(FIPS140.useCompliantAlgorithms() && (bindName == null || bindPassword == null || bindPassword.length() < 14)) {
```
gbhat618, line 153:
```suggestion
        if(FIPS140.useCompliantAlgorithms() && (bindPassword == null || bindPassword.length() < 14)) {
```

I contemplated about keeping the `bindName == null` check and making it part of the `IllegalArgumentException` but instead of combining it with the password check perhaps better to put another check for it.

gbhat618: given that `bindName` is only then used with `fixEmpty(bindName)` in this constructor code (other methods have appropriate checks), there is no need to introduce the `null` check again.

the `fixEmpty` honors null, doesn't result in NPE.

Safe to accept the suggestion removing `bindName == null` in here.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -212,7 +208,7 @@ protected Object readResolve() {
         }
 
         String bindPassword_ = Secret.toString(bindPassword);
-        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword_) < 14) {
+        if(FIPS140.useCompliantAlgorithms() && bindPassword_.length() < 14) {
```
gbhat618, line 211:
```suggestion
        if(FIPS140.useCompliantAlgorithms() && (bindPassword_ == null || bindPassword_.length() < 14)) {
```

returns 0 if null https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#length(java.lang.CharSequence).

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -294,7 +290,7 @@ public ListBoxModel doFillTlsConfigurationItems() {
          */
         @RequirePOST
         public FormValidation doCheckBindPassword(@QueryParameter String bindPassword) {
-            if(FIPS140.useCompliantAlgorithms() && StringUtils.length(bindPassword) < 14) {
+            if(FIPS140.useCompliantAlgorithms() && bindPassword.length() < 14) {
```
gbhat618, line 293:
```suggestion
            if(FIPS140.useCompliantAlgorithms() && (bindPassword == null || bindPassword.length() < 14)) {
```
null check required.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
```diff
@@ -374,51 +370,31 @@ public FormValidation doValidateTest(@QueryParameter(fixEmpty = true) String nam
...
-                        }
-                        return FormValidation.error(e, "Bad bind username or password");
-                    } catch (javax.naming.AuthenticationException e) {
-                        return FormValidation.error(e, "Bad bind username or password");
-                    } catch (ServiceUnavailableException e) {
-                        return FormValidation.error(e, "Domain Controller is reachable but the service on the specified port is not reachable");
-                    } catch (Exception e) {
-                        return FormValidation.error(e, e.getMessage());
-                    }
-                } else {
```
gbhat618, line 404: this was a dead `else` block, when `bindName == null` the code wouldn't have reached here, instead returned at L354.

gbhat618 (approved):
lgtm with 2 minor comments.

gbhat618:
Hi @alecharp, need help check some of the comments. Later will ask help for merge.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
alecharp: `Secret.toString` doesn't return `null`.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
alecharp: If `bindName` is `null` then the exception thrown next is not correct.

src/main/java/hudson/plugins/active_directory/ActiveDirectoryDomain.java
alecharp: No, the `bindPassword` coming from the form is never `null`.

gbhat618:
> need help check some of the comments. Later will ask help for merge.

all comments addressed with no change - good. then we can merge it.

Hi @jtnord could you help to move this PR, already reviewed, looks good. ty.

